PR #11, opened 2020-09-11
Implement basic _include and _revinclude

carvantes:
Description of changes:

_include and _revinclude allow to fetch related resources as part of search results.

For example:
```
GET [base]/MedicationRequest?_include=MedicationRequest:patient
GET [base]/MedicationRequest?_revinclude=Provenance:target
```
The first search requests all matching MedicationRequests, to include any patient that the medication prescriptions in the result set refer to. The second search requests all matching prescriptions, return all the provenance resources that refer to them. 

`_include` and `_revinclude` also support wildcards and have an `:iterate` variant, but those features are out of the scope of this PR

Also added missing unit tests. Elasticsearch queries are tough to validate without actually running them against an Elasticsearch instance so we are not making specific assertions on them but rather using snapshot testing.
Snapshots let us see what queries are run for the different search requests in test cases and serve as nice regression tests. 


src/__tests__/elasticSearchService.test.ts
```diff
@@ -0,0 +1,195 @@
...
+jest.mock('../elasticSearch');
+
+const FILTER_RULES_FOR_ACTIVE_RESOURCES = [{ match: { someFieldThatTellsIfTheResourceIsActive: 'AVAILABLE' } }];
+
+describe('typeSearch', () => {
+    beforeEach(() => {
+        jest.resetAllMocks();
+    });
+    describe('query snapshots for simple queryParams', () => {
+        each([
```
rsmayda: @nguyen102 this seems like a cleaner approach for 'each' -- thoughts?

src/__tests__/elasticSearchService.test.ts
```diff
@@ -0,0 +1,195 @@
...
+            };
+            (ElasticSearch.search as jest.Mock).mockResolvedValue(fakeSearchResult);
+            const es = new ElasticSearchService(FILTER_RULES_FOR_ACTIVE_RESOURCES);
+            await es.typeSearch({
+                resourceType: 'Patient',
+                baseUrl: 'https://base-url.com',
+                queryParams,
+            });
+
+            expect((ElasticSearch.search as jest.Mock).mock.calls).toMatchSnapshot();
```
rsmayda: What are we testing here? that the search result returned?

src/elasticSearchService.ts
```diff
@@ -17,6 +19,14 @@ import {
 import { ElasticSearch } from './elasticSearch';
 import { DEFAULT_SEARCH_RESULTS_PER_PAGE, SEARCH_PAGINATION_PARAMS } from './constants';
 
+const NON_SEARCHABLE_FIELDS = [
+    SEARCH_PAGINATION_PARAMS.PAGES_OFFSET,
```
rsmayda, line 23: Does pagination still work when using rev/includes?

rsmayda: FYI from FHIR spec: 
`When search results are paged, each page of search results should include the matching includes for the resources in each page, so that each page stands alone as a coherent package.`

carvantes: yes, pagination works with rev/include. However, rev/include brings in additional resources that are outside the pagination count limit.

i.e. `[base]/MedicationRequest?_include=MedicationRequest:patient&_count=10` can return more than 10 resources. 10 MedicationRequest + all the related patients

src/__tests__/elasticSearchService.test.ts
carvantes: It's testing that the executed ES queries match the ones in the snapshot. you can see the snapshots at `src/__tests__/__snapshots__/elasticSearchService.test.ts.snap`

src/elasticSearchService.ts
rsmayda: Is that expected as per FHIR spec?

src/__tests__/elasticSearchService.test.ts
nguyen102: nice! I like this syntax

src/__tests__/elasticSearchService.test.ts
```diff
@@ -0,0 +1,195 @@
...
+                {
+                    _count: 10,
+                    _getpagesoffset: 2,
+                    id: '11111111-1111-1111-1111-111111111111',
+                    gender: 'female',
+                    name: 'Emily',
+                    _format: 'json',
+                },
+            ],
+        ]).test('queryParams=%j', async (queryParams: any) => {
```
nguyen102: is the `j` supposed to be here? In `queryParams=%j`

src/elasticSearchService.ts
```diff
@@ -142,24 +131,185 @@ export class ElasticSearchService implements Search {
...
+                    .filter(reference => RELATIVE_URL_REGEX.test(reference))
+                    .map(relativeUrl => {
+                        const [resourceType, id] = relativeUrl.split('/');
+                        return { resourceType, id };
+                    })
+                    .filter(({ resourceType }) => !targetResourceType || targetResourceType === resourceType);
+            }),
+        );
+
+        const idsByResourceType = mapValues(
```
nguyen102: Nitpick: Is this a map of resourceType to ids? We've been using the format `keyToValue`. Rename the variable `idsByResourceType` => `resourceTypeToIds`

src/elasticSearchService.ts
```diff
@@ -142,24 +131,185 @@ export class ElasticSearchService implements Search {
...
+        searchEntries: SearchEntry[],
+        request: TypeSearchRequest,
+    ): Promise<SearchEntry[]> {
+        const { queryParams, baseUrl } = request;
+        if (!queryParams._include) {
+            return [];
+        }
+        const includes = this.getParamAsArray(queryParams._include);
+
+        const resourcesToInclude = flatten(
```
nguyen102: What is an example value for this variable? Is this a map of ressources to the include variable?

src/__tests__/elasticSearchService.test.ts
carvantes: yes, it's used for formatting. `%j` is for json so the text name looks as follows:
```
 typeSearch
    query snapshots for simple queryParams
      ✓ queryParams={} (5ms)
      ✓ queryParams={"_count":10,"_getpagesoffset":2} (2ms)

```

https://www.npmjs.com/package/jest-each

src/elasticSearchService.ts
carvantes: It has type:
```
{resourceType: string, id: string}[]
```

src/elasticSearchService.ts
carvantes: done

src/elasticSearchService.ts
carvantes: I added the type explicitly. It makes it much more readable

nguyen102: Ah ok. Thanks for the clarification. I think the naming of this variable is to indicate which resources to search for for `searchIncludes`, but it's a little confusing since when I first read it, I thought it was a map with key of resource and value of id. We've been using the syntax for variable name of `keyToValue`. Could we rename this to something like `resourceTypeToIdForIncludedResources`. It's kind of a long variable name, but I'm open to suggestions for other ideas.

src/elasticSearchService.ts
```diff
@@ -142,24 +131,185 @@ export class ElasticSearchService implements Search {
...
+
+    // eslint-disable-next-line class-methods-use-this
+    private getParamAsArray(param: any): string[] {
+        if (!param) {
+            return [];
+        }
+        return Array.isArray(param) ? (uniq(param) as string[]) : [param as string];
+    }
+
+    private async processSearchIncludes(
```
nguyen102, line 194: I think it might be helpful to move some of these private methods to a class like `elasticSearchServiceHelper` and then make those methods `public`. That way this will keep `elasticSearchService` smaller, and also allow us to unit tests these methods. 

There's a lot of variable mapping in this method, and it makes it a little hard to follow what each variable are as they're being transformed. Being able to unit tests this method and maybe even components of this methods would make it easier to understand what is going on.

src/elasticSearchService.ts
carvantes: yes,
> Note that the Bundle.total only include the total number of matching resources. It does not count extra resources such as OperationOutcome or included resources that may also be returned

src/elasticSearchService.ts
carvantes: I split it into functions and moved it to a separate file.

src/elasticSearchService.ts
carvantes: Renamed to simply `resourceReferences`. Also I did some refactoring so this var is now on a smaller function so it's easier to understand.

btw, This is var involves objects, but they are not  as a map, so using `To` like in `resourceTypeToIdForIncludedResources ` does not really apply.
i.e:
```
[
  { resourceType: 'Practitioner', id: '111' },
  { resourceType: 'Organization', id: '222' },
]
```

src/searchInclusions.ts
```diff
@@ -0,0 +1,175 @@
...
+export const inclusionParameterFromString = (s: string): InclusionSearchParameter | null => {
+    const INCLUSION_PARAM_REGEX = /^(?<sourceResource>[A-Za-z]+):(?<searchParameter>[A-Za-z.]+)(?::(?<targetResourceType>[A-Za-z]+))?$/;
+    const match = s.match(INCLUSION_PARAM_REGEX);
+    if (match === null) {
+        // Malformed inclusion search parameters are ignored. No exception is thrown.
+        // This allows the regular search to complete successfully
+        console.log(`Ignoring invalid include/revinclude search parameter: ${s}`);
+        return null;
+    }
+    const { sourceResource, searchParameter, targetResourceType } = match.groups!;
```
nguyen102, line 28: Ooo, interesting. What does `!` in `match.groups` do?

carvantes: It's called Non-null assertion operator, https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator

It is safe to use it here since we've already tested that `match` is not null and we know for sure that `groups` is not null because the regex has capture groups.

src/__tests__/SearchInclusions.test.ts
```diff
@@ -0,0 +1,169 @@
...
+                sourceResource: 'Patient',
+                searchParameter: 'field',
+                targetResourceType: 'OtherResource',
+            };
+            expect(inclusionParameterFromString(input)).toEqual(expected);
+        });
+    });
+});
+
+describe('getInclusionParametersFromQueryParams', () => {
```
nguyen102: Thanks for adding these tests! They make it a lot easier to understand what the input and output of the functions under test are.

nguyen102 (changes requested):
Overall looks good. Just one nitpick. Our current test convention is to put the test file next to the file being tested. For example. https://github.com/awslabs/fhir-works-on-aws-routing/tree/mainline/src/router/handlers

So if we can just keep the consistent with the tests here..

carvantes:
> Overall looks good. Just one nitpick. Our current test convention is to put the test file next to the file being tested. For example. https://github.com/awslabs/fhir-works-on-aws-routing/tree/mainline/src/router/handlers
> 
> So if we can just keep the consistent with the tests here.

Done

nguyen102 (approved):

package.json
```diff
@@ -61,7 +64,7 @@
     "transform": {
       "\\.(ts)$": "ts-jest"
     },
-    "testRegex": ".test.ts"
+    "testRegex": "(/__tests__/.*|(\\.|/)(test|spec))\\.tsx?$"
```
rsmayda: may want to revert?

src/SearchInclusions.test.ts
```diff
@@ -0,0 +1,169 @@
+import each from 'jest-each';
```
rsmayda: nit: filename between test and non test has a casing difference. make this file name with a lowercase 's'

src/elasticSearchService.ts
```diff
@@ -142,24 +131,107 @@ export class ElasticSearchService implements Search {
                 );
             }
 
+            const [includedResources, revincludedResources] = await Promise.all([
+                this.processSearchIncludes(result.entries, request),
+                this.processSearchRevIncludes(result.entries, request),
+            ]);
```
rsmayda, line 137: It looks like in this method we are querying ES for includes/revincludes even when it is not in the query -- is that right? Actually looking at searchInclusions.ts you have a helper method that will return an empty array which in turn will make no queries ran.. 

^^ Please confirm this :)

package.json
carvantes: Done. Although it did require a minor change to `.test.ts$` in order to exclude test snapshot files.

src/SearchInclusions.test.ts
carvantes: done

src/elasticSearchService.ts
carvantes: That's correct. 

These methods always run but may end up not executing any ES queries and just returning an empty array.

rsmayda (approved):

PR #12, opened 2020-09-18
feat: Add support for wildcards in _include & _revinclude parameters

carvantes:
Description of changes:

Add support for wildcards in _revinclude and _include parameters

For example,
```
{{API_URL}}/MedicationRequest?_revinclude=*
```
fetches ALL resources that reference the MedicationRequests. 
https://www.hl7.org/fhir/search.html#include

---

The basic idea of the implementation is to add a step that "expands" the wildcard parameter into all the possible search parameters. 

To achieve this, we use the files at `src/schema` that describe all the references between FHIR resources. These files are derived from the FHIR Definitions from https://www.hl7.org/fhir/downloads.html. I added `scripts/generateResourceReferenceMatrixFile.js` to facilitate generating new files for other FHIR versions.

References are different across FHIR versions, `ElasticSearchService` now accept `fhirVersion` as a constructor parameter so that it can pick the correct resource references file to resolve the wildcard inclusions.

It's worth noting that _revinclude=* is expensive since some resource types are referenced by many other resource types. And 1 query is needed per resource type since they have their own index. For example `/Patient?_revinclude=*` will issue 70+ ES queries

Note:
Also fixed a couple minor bugs:
- Properly handle fields that are an array of references. Before they were ignored.
- Dedupe references before building Include queries. This wasn't breaking anything, but duplicate id terms in the query serve no purpose.


scripts/generateResourceReferenceMatrixFile.js
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+const fs = require('fs');
+const stringify = require('json-stringify-pretty-compact');
+
+const data = JSON.parse(fs.readFileSync('profiles-resources.json'));
+
+const { fhirVersion } = data.entry[0].resource;
+
+if (fhirVersion.startsWith('4')) {
```
rsmayda, line 20: Supported FHIR versions by FHIR Works is a type: https://github.com/awslabs/fhir-works-on-aws-interface/blob/mainline/src/constants.ts#L78

I can see the argument of keeping it generic, what are your thoughts

scripts/generateResourceReferenceMatrixFile.js
```diff
@@ -0,0 +1,62 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+/*
+This scripts generates the JSON files at src/schema. Before running the script download the JSON FHIR definition package and copy
+the profiles-resources.json file on this directory.
```
rsmayda:
```suggestion
the profiles-resources.json file into this directory.
```

scripts/generateResourceReferenceMatrixFile.js
```diff
@@ -0,0 +1,62 @@
...
+const stringify = require('json-stringify-pretty-compact');
+
+const data = JSON.parse(fs.readFileSync('profiles-resources.json'));
+
+const { fhirVersion } = data.entry[0].resource;
+
+if (fhirVersion.startsWith('4')) {
+    const r = data.entry
+        .filter(x => x.resource.baseDefinition === 'http://hl7.org/fhir/StructureDefinition/DomainResource')
+        .flatMap(x => {
```
rsmayda: better param names `r, x, a, b, c, el`

src/searchInclusions.ts
```diff
@@ -27,24 +33,59 @@ export const inclusionParameterFromString = (s: string): InclusionSearchParamete
...
     queryParams: any,
-): TypedInclusionParameter[] => {
+): (InclusionSearchParameter | WildcardInclusionSearchParameter)[] => {
     const queryParam = queryParams?.[includeType];
     if (!queryParam) {
         return [];
     }
     if (Array.isArray(queryParam)) {
         return uniq(queryParam)
             .map(x => inclusionParameterFromString(x))
```
rsmayda: better param name, perhaps `param`

src/searchInclusions.ts
```diff
@@ -27,24 +33,59 @@ export const inclusionParameterFromString = (s: string): InclusionSearchParamete
...
+            targetResourceType: targetResourceType === 'Resource' ? undefined : targetResourceType,
+        }));
+};
+
 export const getInclusionParametersFromQueryParams = (
     includeType: '_include' | '_revinclude',
     queryParams: any,
-): TypedInclusionParameter[] => {
+): (InclusionSearchParameter | WildcardInclusionSearchParameter)[] => {
     const queryParam = queryParams?.[includeType];
```
rsmayda, line 82:
```suggestion
    const inclusionParams = queryParams?.[includeType];
```

src/searchInclusions.ts
```diff
@@ -53,19 +94,25 @@ export const getInclusionParametersFromQueryParams = (
...
         return resources
-            .map(resource => resource[include.searchParameter]?.reference)
+            .map(resource => get(resource, `${include.searchParameter}`) as any)
+            .flatMap(valueAtPath => {
+                if (Array.isArray(valueAtPath)) {
+                    return valueAtPath.map(v => get(v, 'reference'));
+                }
+                return [get(valueAtPath, 'reference')];
+            })
             .filter((x): x is string => typeof x === 'string')
```
rsmayda: better param for x will make this chaining easier to read

rsmayda (commented):
You mention '*' can be slow -- Out of curiosity how 'slow' is a wild card search?

scripts/generateResourceReferenceMatrixFile.js
```diff
@@ -0,0 +1,62 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+/*
+This scripts generates the JSON files at src/schema. Before running the script download the JSON FHIR definition package and copy
```
nguyen102, line 7: I think the information here would need to be placed in the `deployment` package `INSTALLATION.md` file. Because what would happen if a user does a wild card search, but doesn't have the `profiles-resource.json` file setup? Would this error out?

Also should we pregenerate the matrix file for FHIR R4.0.1 and STU3? That way if the user is already using one of our supported FHIR versions, they won't have to do the manual step of generating these matrix files?

scripts/generateResourceReferenceMatrixFile.js
```diff
@@ -0,0 +1,62 @@
...
+the profiles-resources.json file into this directory.
+
+You can download the latest FHIR definition from https://www.hl7.org/fhir/downloads.html or find older FHIR versions at http://hl7.org/fhir/directory.html
+ */
+
+const fs = require('fs');
```
nguyen102, line 13: I'm debating whether we need to write a test for this script. In the past we haven't written tests for scripts, but without a test for this script, I think it'll be very hard to make any changes in this script in the future and feel safe about the changes.

Would it be hard to write a test for this script? Maybe using a subset of the `profile-resources.json` file?

nguyen102: Oh nice, I see the matrix files are already included.

src/schema/fhirResourceReferencesMatrix.v3.0.1.json
```diff
@@ -0,0 +1,1084 @@
...
+  ["Observation", "related.target", "QuestionnaireResponse"],
+  ["Observation", "related.target", "Sequence"],
+  ["OperationDefinition", "base", "OperationDefinition"],
+  ["OperationDefinition", "parameter.profile", "StructureDefinition"],
+  ["OperationDefinition", "parameter.binding.valueSet", "ValueSet"],
+  ["Organization", "partOf", "Organization"],
+  ["Organization", "endpoint", "Endpoint"],
+  ["Patient", "contact.organization", "Organization"],
+  ["Patient", "generalPractitioner", "Organization"],
+  ["Patient", "generalPractitioner", "Practitioner"],
```
nguyen102, line 808: Should there be a row 
` ["Patient", "generalPractitioner", "PractitionerRole"],`?

From the `profile-resources.json`
```
     {
          "id" : "Patient.generalPractitioner",
          "path" : "Patient.generalPractitioner",
          "short" : "Patient's nominated primary care provider",
          "definition" : "Patient's nominated care provider.",
          "comment" : "This may be the primary care provider (in a GP context), or it may be a patient nominated care manager in a community/disability setting, or even organization that will provide people to perform the care provider roles.  It is not to be used to record Care Teams, these should be in a CareTeam resource that may be linked to the CarePlan or EpisodeOfCare resources.\nMultiple GPs may be recorded against the patient for various reasons, such as a student that has his home GP listed along with the GP at university during the school semesters, or a \"fly-in/fly-out\" worker that has the onsite GP also included with his home GP to remain aware of medical issues.\n\nJurisdictions may decide that they can profile this down to 1 if desired, or 1 per type.",
          "alias" : ["careProvider"],
          "min" : 0,
          "max" : "*",
          "base" : {
            "path" : "Patient.generalPractitioner",
            "min" : 0,
            "max" : "*"
          },
          "type" : [{
            "code" : "Reference",
            "targetProfile" : ["http://hl7.org/fhir/StructureDefinition/Organization",
            "http://hl7.org/fhir/StructureDefinition/Practitioner",
            "http://hl7.org/fhir/StructureDefinition/PractitionerRole"]
          }],
```
I downloaded the json from this link
https://www.hl7.org/fhir/definitions.json.zip

scripts/generateResourceReferenceMatrixFile.js
carvantes: This is a `.js` file so no types. 

It's convenient to not use typescript here or lock down the accepted FHIR versions since this script is intended for developers to generate files if other FHIR versions need to be supported. 
Just download the FHIR resources for a different version, run `node scripts/generateResourceReferenceMatrixFile.js` and commit the generated files.

It is not part of `src` so it's not tsc compiled and does not end up in the npm package 

Although this comment made me realize that I should be using the `FhirVersion` type on other parts of the code.

scripts/generateResourceReferenceMatrixFile.js
carvantes: updated

src/searchInclusions.ts
carvantes: updated

src/searchInclusions.ts
carvantes: updated

src/searchInclusions.ts
carvantes: I'm keeping the name as is. 

I find `queryParam` clearer. At this point we are only picking a value from the queryParams object, but it could be anything. It may be invalid or even undefined. It doesn't have the `InclusionSearchParameter` type just yet.

later on more specific names like `inclusionParameter` are used for vars of type `InclusionSearchParameter`

src/schema/fhirResourceReferencesMatrix.v3.0.1.json
carvantes: You downloaded the json for the latest version(4.0.1) but you are looking at the 3.0.1 generated file

`["Patient", "generalPractitioner", "PractitionerRole"],` can be found on `fhirResourceReferencesMatrix.v4.0.1.json`

scripts/generateResourceReferenceMatrixFile.js
carvantes: The generated files are included in this commit.

This script is intended for developers only to facilitate generating source code for other FHIR versions. Customers don't need to run it.

https://github.com/awslabs/fhir-works-on-aws-search-es/pull/12#discussion_r491056557

scripts/generateResourceReferenceMatrixFile.js
carvantes: I don't intend to write tests for this script. 

There's very little reason to modify this script in the future. Plus, committing the generated files sort of serves as a snapshot test. i.e. if in the future you modify the script, and re run it for v4.0.1 and see a different output, you may have introduced a bug.

nguyen102: Yeah, it is unlikely for the `profile-resource.json` specs to change. So that would make it unlikely for this script to need changes. I'm ok with not adding tests for now.

src/schema/fhirResourceReferencesMatrix.v3.0.1.json
nguyen102: Ah, thanks for the clarification.

scripts/generateResourceReferenceMatrixFile.js
nguyen102: Since the matrix files are already generated, that's good. We don't need any special instructions for a consumer of the `deployment` package to make changes to this package. 

However, should we add instructions in the README of this package, to let user knows about this script and how to use it? Otherwise, they wouldn't know what this script is or that it's even available unless they open this script up.

carvantes: I'd rather not. 

The README is directed at consumers of the npm package, but this script is not part of the npm package. The same README also shows up at the NPM website.

We could add some docs in a different `.md` file, but I don't think it's necessary since the use case for this script is very narrow(new FHIR versions don't come out often) and the script is self documented with comments.

nguyen102: Ah, ok, if the README shows up on the NPM package page, then we should probably not add instructions there. I still think we should document this script somewhere. Although it's not an approval blocking change though.

nguyen102 (approved):

carvantes:
@rsmayda 
> You mention '*' can be slow -- Out of curiosity how 'slow' is a wild card search?

It's not necessarily slow, all revinclude queries run in parallel.

load test, 10 TPS during 3 min:
Screen Shot 2020-09-21 at 18 00 04

Granted, this test used a very small dataset w most queries returning empty results.

btw, I was originally using the search api for each query, but I switched to the multi search api. This way even if there are 70+ queries, a single request to ES is made(that made a big difference)

PR #13, opened 2020-09-23
feat: Implement _include:iterate & _revinclude:iterate

carvantes:
Description of changes:

These PR adds support for the `_include:iterate` and `_revinclude:iterate` search parameters

From the FHIR spec:

The inclusion process can be iterative, if (and only if) the modifier :iterate is included. For example, this search returns all Medication Request resources and their prescribing Practitioner Resources for the matching Medication Dispense resources:

```
GET [base]/MedicationDispense?_include=MedicationDispense:prescription&_include:iterate=MedicationRequest:performer&criteria...
```
This technique applies to circular relationships as well. For example, the first of these two searches includes any related observations to the target relationships, but only those directly related. The second search asks for the _include based on related parameter to be executed iteratively, so it will retrieve observations that are directly related, and also any related observations to any other included observation.

```
GET [base]/Observation?_include=Observation:related-target&criteria...
GET [base]/Observation?_include:iterate=Observation:related-target&criteria...
```

---
Evaluating `:iterate` inclusions is the last step of search since it operates on top of the resources found by regular search criteria and the resources found by regular _include/_revinclude

The basic idea is to run the already existing inclusion process iteratively, feeding the resources found by one iteration as the starting resources for the next one. Iterations stop once either:
- No more resources are found. This means that the inclusion process is complete
- `MAX_INCLUDE_ITERATIVE_DEPTH` is reached. This protects us from returning way too many results.
> Using iterative wildcards inclusions might lead to the retrieval of the full patient's record, or even more than that: resources are organized into an interlinked network and broad _include paths may eventually traverse all possible paths on the server.
It is at the server's discretion how deep to iteratively evaluate the inclusions. Servers are expected to limit the number of iterations done to an appropriate level and are not obliged to honor requests to include additional resources in the search results.

Other changes:

- include & revinclude queries are now sent together in a single call to ES. Less calls to ES is better.
- Updated several SearchInclusions functions to operate on multiple resource types at once. Before we could assume that all the resources matched the `requestResourceType`. This is no longer true for `:iterate`

### DIFF
--- a/src/elasticSearchService.ts
+++ b/src/elasticSearchService.ts
@@ -26,7 +26,11 @@ const NON_SEARCHABLE_FIELDS = [
     '_format',
     '_include',
     '_revinclude',
+    '_include:iterate',
+    '_revinclude:iterate',
 ];
+
+const MAX_INCLUDE_ITERATIVE_DEPTH = 5;
 
 // eslint-disable-next-line import/prefer-default-export
 export class ElasticSearchService implements Search {
@@ -137,11 +141,12 @@ export class ElasticSearchService implements Search {
                 );
             }
 
-            const [includedResources, revincludedResources] = await Promise.all([
-                this.processSearchIncludes(result.entries, request),
-                this.processSearchRevIncludes(result.entries, request),
-            ]);
-            result.entries.push(...includedResources, ...revincludedResources);
+            const includedResources = await this.processSearchInclusions(result.entries, request);
+            result.entries.push(...includedResources);
+
+            const iterativelyIncludedResources = await this.processIterativeSearchInclusions(result.entries, request);
+            result.entries.push(...iterativelyIncludedResources);
+
             return { result };
         } catch (error) {
             console.error(error);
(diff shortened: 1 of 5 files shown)


src/elasticSearchService.ts
```diff
@@ -26,8 +26,12 @@ const NON_SEARCHABLE_FIELDS = [
     '_format',
     '_include',
     '_revinclude',
+    '_include:iterate',
+    '_revinclude:iterate',
 ];
 
+const MAX_INCLUDE_ITERATIVE_DEPTH = 5;
```
rsmayda, line 33: With Lambda upper limit of being response size of 6mb and using an average resource size of 100kb that means we can return ~6mb/100kb = 60 resources;

Do you think the iterative depth is too large? or perhaps the depth could be more dynamic (ie when the response size reaches >50 entries)

src/elasticSearchService.ts
```diff
@@ -233,36 +238,71 @@ export class ElasticSearchService implements Search {
...
+            searchEntries.map(searchEntry => searchEntry.resource.id),
         );
+        const resourceIdsWithInclusionsAlreadyResolved: Set<string> = new Set();
 
-        const { hits } = await this.executeQueries(searchQueries);
-        return this.hitsToSearchEntries({ hits, baseUrl: request.baseUrl, mode: 'include' });
+        console.log('Iterative inclusion search starts');
+
+        let resourcesToIterate = searchEntries;
+        for (let i = 0; i < MAX_INCLUDE_ITERATIVE_DEPTH; i += 1) {
```
rsmayda, line 278: Out of curiousity does a fhir client know if the results are limited or not (from the FHIR spec it doesn't seem to specify)

rsmayda (commented):
Question around AWS infrastructure limitations

src/elasticSearchService.ts
carvantes: From what I've read, no there's no way to tell. 

Although there's an optional way of communicating errors described at: https://www.hl7.org/fhir/search.html#errors
> In such cases, the search process MAY include an OperationOutcome in the search set that contains additional hints and warnings about the search process. This is included in the search results as an entry with search mode = outcome. Clients can use this information to improve future searches. If, for example, a client performed the following search:

>GET [base]/Observation?patient.identifier=http://example.com/fhir/identifier/mrn|123456

>and there is no patient for MRN 123456, the server would return a bundle with a warning.

The guidance is for errors in general, not just for include stuff. And it's completely optional. I guess it'd be possible to return a bundle with a warning OperationOutcome telling that the included results are not complete due to a limit being exceeded.

src/elasticSearchService.ts
carvantes: 5 is just an educated guess.

100kb may be too big for an avg resource size.

running `http://hapi.fhir.org/baseR4/Patient?_include:iterate=*&_revinclude:iterate=*&_count=500` on the HAPI server returns a total of 5048 resources of 27 different resource types with a response size of 7.5MB. That averages **1.49KB per resource**

From that, 6MB should fit around 4k resources.

Resources generated from Synthea give a similar number. The entire record of a Patient generates 1.2k resources averaging **1.49KB per resource** 

I think that starting with a depth limit of 5 is ok. It's something that we can tweak in the future based on customer feedback. 
We can work later on a task focused on ensuring that this lambda does not exceed the max payload size(we don't even have a max page size defined right now)

---
btw `http://hapi.fhir.org/baseR4/Patient?_include:iterate=*&_revinclude:iterate=*&_count=500` is a crazy query. I tend to think that customers issue much more specific queries all the time. I need more customer feedback to validate my assumptions though.

wildcards are known to be slow and are frowned upon by FHIR implementations:
>You can include all resources referenced from the search result using *. This is considered bad practice because it's too implicit. This feature is only implemented for conformance with the FHIR specification. Please avoid using it! 

and some implementations outright not support them in iterations:
>_include=* could not be used as part of chained (rev)includes!

https://docs.aidbox.app/basic-concepts/search-1/_include-and-_revinclude

rsmayda: Thanks for this great research -- we should add this documentation somewhere. knowing the 'average' resource size will be good for future designs

src/elasticSearchService.ts
rsmayda: Thanks for checking I say we leave it as is and adjust if customer feedback deems it so

rsmayda (approved):

PR #14, opened 2020-09-24
feat: Narrow down included resources to the allowed resource types

carvantes:
Description of changes:

Narrow down the search results to the allowed resource types as described in https://github.com/awslabs/fhir-works-on-aws-interface/pull/16

**Note**: Need to publish the updated interface to npm first. Until then, checks are failing.


src/elasticSearchService.test.ts
```diff
@@ -63,6 +76,7 @@ describe('typeSearch', () => {
                 resourceType: 'Patient',
                 baseUrl: 'https://base-url.com',
                 queryParams,
+                allowedResourceTypes: ALLOWED_RESOURCE_TYPES,
```
rsmayda, line 79: Could we have 1 test with this being an empty array?

src/elasticSearchService.ts
```diff
@@ -258,7 +258,13 @@ export class ElasticSearchService implements Search {
...
             iterative,
         );
-        const { hits } = await this.executeQueries([...includeSearchQueries, ...revIncludeSearchQueries]);
+
+        const lowerCaseAllowedResourceTypes = new Set(request.allowedResourceTypes.map(r => r.toLowerCase()));
+        const allowedInclusionQueries = [...includeSearchQueries, ...revIncludeSearchQueries].filter(query =>
+            lowerCaseAllowedResourceTypes.has(query.index),
+        );
+
+        const { hits } = await this.executeQueries(allowedInclusionQueries);
```
rsmayda, line 267: FYI just verified we are not breaking from the FHIR standard :)
>... Systems responding to queries might suppress records or may adjust the content of resource instances to exclude data elements the querying system is not permitted to see.

https://www.hl7.org/fhir/updates.html#7.14.3

carvantes: There's more specific guidance on https://www.hl7.org/fhir/security.html

> _include and _revinclude search parameters allow client to request related resources. A security implementation must determine if the client has access to the included resources.

src/elasticSearchService.test.ts
carvantes: Added a test for it (Not with an empty array though but rather with a single type on it. At the very least you are always allowed to search for the `TypeSearchRequest.resourceType `)

rsmayda (approved):

PR #20, opened 2020-10-13
refactor: Don't start iterative inclusion search if there are no iterative inclusion parameters

nguyen102:
Description of changes:
Don't start iterative inclusion search if there are no iterative inclusion parameters

### DIFF
--- a/src/elasticSearchService.ts
+++ b/src/elasticSearchService.ts
@@ -275,7 +275,7 @@ export class ElasticSearchService implements Search {
     ): Promise<SearchEntry[]> {
         if (
             !ITERATIVE_INCLUSION_PARAMETERS.some(param => {
-                return Object.keys(request.queryParams).includes(param);
+                return Object.prototype.hasOwnProperty.call(request.queryParams, param);
             })
         ) {
             return [];
--- a/src/elasticSearchService.ts
+++ b/src/elasticSearchService.ts
@@ -21,14 +21,15 @@ import { DEFAULT_SEARCH_RESULTS_PER_PAGE, SEARCH_PAGINATION_PARAMS } from './con
 import { buildIncludeQueries, buildRevIncludeQueries } from './searchInclusions';
 import { getDocumentField } from './searchParametersMapping';
 
+const ITERATIVE_INCLUSION_PARAMETERS = ['_include:iterate', '_revinclude:iterate'];
+
 const NON_SEARCHABLE_PARAMETERS = [
     SEARCH_PAGINATION_PARAMS.PAGES_OFFSET,
     SEARCH_PAGINATION_PARAMS.COUNT,
     '_format',
     '_include',
     '_revinclude',
-    '_include:iterate',
-    '_revinclude:iterate',
+    ...ITERATIVE_INCLUSION_PARAMETERS,
 ];
 
 const MAX_INCLUDE_ITERATIVE_DEPTH = 5;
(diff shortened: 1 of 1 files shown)


src/elasticSearchService.ts
```diff
@@ -272,6 +273,13 @@ export class ElasticSearchService implements Search {
         searchEntries: SearchEntry[],
         request: TypeSearchRequest,
     ): Promise<SearchEntry[]> {
+        if (
+            Object.keys(request.queryParams).filter(searchParam => {
+                return ITERATIVE_INCLUSION_PARAMETERS.includes(searchParam);
+            }).length === 0
```
rsmayda: nitpick: I think `some` is clearer:  `arr1.some(r=> arr2.includes(r))`

nguyen102: updated

src/elasticSearchService.ts
```diff
@@ -272,6 +273,13 @@ export class ElasticSearchService implements Search {
         searchEntries: SearchEntry[],
         request: TypeSearchRequest,
     ): Promise<SearchEntry[]> {
+        if (
+            !ITERATIVE_INCLUSION_PARAMETERS.some(param => {
+                return Object.keys(request.queryParams).includes(param);
```
carvantes: Just access queryParams by key instead of doing .includes on the array of keys

carvantes (changes requested):

src/elasticSearchService.ts
```diff
@@ -272,6 +273,13 @@ export class ElasticSearchService implements Search {
         searchEntries: SearchEntry[],
         request: TypeSearchRequest,
     ): Promise<SearchEntry[]> {
+        if (
+            !ITERATIVE_INCLUSION_PARAMETERS.some(param => {
+                return Object.prototype.hasOwnProperty.call(request.queryParams, param);
```
nguyen102: Using `Object.prototype.hasOwnProperty.call` as per eslint's suggestion https://eslint.org/docs/rules/no-prototype-builtins

src/elasticSearchService.ts
nguyen102: updated

src/elasticSearchService.ts
carvantes: why are we using `Object.prototype.hasOwnProperty` instead of just checking `queryParams[param]` ?

nguyen102: Ah good question. 

Using `queryParams[param]` could introduce a subtle bug. Example shown below
```
var obj = { key: undefined };
obj["key"] !== undefined // false, but the key exists!

Comment (nguyen102) [reply src/elasticSearchService.ts]: Ah good question. 

Using `queryParams[param]` could introduce a subtle bug. Example shown below
```
var obj = { key: undefined };
obj["key"] !== undefined // false, but the key exists!
```
Granted it might not be possible for this particular variable, `queryParams[param]` to be explicitly defined as `undefined` when a user make a search request.  However, I don't think it hurts to use `hasOwnProperty` to be extra safe. 

More details in this [link](https://stackoverflow.com/a/1098955)

carvantes: Code is correct, but find it misleading:
- It implies that this function cares about the difference between a missing key and a key with value of undefined. It doesn't, they are the same for this function, there's no subtle bug here.
- It implies that we are giving special treatment to inherited properties. We aren't. This isn't too relevant since as you mention `queryparams` is expected to be a simple object generated by Express. Even then, if we consider a weird case where queryparams inherited the `_include:iterate` property, we probably should be reading that inherited property.

nguyen102: updated

carvantes (approved):

PR #28, opened 2020-12-10
feat: Implement date prefix based search

PrimKruskal:
Issue #, if available:
Description of changes:

These PR adds support to search based on dates and its prefixes 

From the FHIR spec:

A date parameter searches on a date/time or period. As is usual for date/time related functionality, while the concepts are relatively straight-forward, there are a number of subtleties involved in ensuring consistent behavior.

The date parameter format is yyyy-mm-ddThh:mm:ss[Z|(+|-)hh:mm] (the standard XML format).

**Prefixes:**
eq | the value for the parameter in the resource is equal to the provided value | the range of the search value fully contains the range of the target value
-- | -- | --
ne | the value for the parameter in the resource is not equal to the provided value | the range of the search value does not fully contain the range of the target value
gt | the value for the parameter in the resource is greater than the provided value | the range above the search value intersects (i.e. overlaps) with the range of the target value
lt | the value for the parameter in the resource is less than the provided value | the range below the search value intersects (i.e. overlaps) with the range of the target value
ge | the value for the parameter in the resource is greater or equal to the provided value | the range above the search value intersects (i.e. overlaps) with the range of the target value, or the range of the search value fully contains the range of the target value
le | the value for the parameter in the resource is less or equal to the provided value | the range below the search value intersects (i.e. overlaps) with the range of the target value or the range of the search value fully contains the range of the target value
sa | the value for the parameter in the resource starts after the provided value | the range of the search value does not overlap with the range of the target value, and the range above the search value contains the range of the target value
eb | the value for the parameter in the resource ends before the provided value | the range of the search value does overlap not with the range of the target value, and the range below the search value contains the range of the target value

Other changes 
ap - Approximation changes is not yet implemented. 

### DIFF
--- a/src/__snapshots__/elasticSearchService.test.ts.snap
+++ b/src/__snapshots__/elasticSearchService.test.ts.snap
@@ -110,7 +110,15 @@ Array [
       "body": Object {
         "query": Object {
           "bool": Object {
-            "filter": Array [],
+            "filter": Array [
+              Array [
+                Object {
+                  "match": Object {
+                    "someFieldThatTellsIfTheResourceIsActive": "AVAILABLE",
+                  },
+                },
+              ],
+            ],
             "must": Array [],
             "must_not": Array [],
           },
@@ -165,7 +173,15 @@ Array [
       "body": Object {
         "query": Object {
           "bool": Object {
-            "filter": Array [],
+            "filter": Array [
+              Array [
+                Object {
+                  "match": Object {
+                    "someFieldThatTellsIfTheResourceIsActive": "AVAILABLE",
+                  },
+                },
+              ],
+            ],
             "must": Array [],
             "must_not": Array [],
           },
@@ -188,7 +204,15 @@ Array [
       "body": Object {
         "query": Object {
           "bool": Object {
-            "filter": Array [],
+            "filter": Array [
+              Array [
+                Object {
+                  "match": Object {
+                    "someFieldThatTellsIfTheResourceIsActive": "AVAILABLE",
+                  },
+                },
+              ],
+            ],
             "must": Array [],
             "must_not": Array [],
           },
@@ -266,7 +290,15 @@ Array [
       "body": Object {
         "query": Object {
           "bool": Object {
-            "filter": Array [],
+            "filter": Array [
+              Array [
+                Object {
+                  "match": Object {
+                    "someFieldThatTellsIfTheResourceIsActive": "AVAILABLE",
+                  },
+                },
+              ],
+            ],
             "must": Array [],
             "must_not": Array [],
           },
@@ -321,7 +353,15 @@ Array [
       "body": Object {
         "query": Object {
           "bool": Object {
-            "filter": Array [],
+            "filter": Array [
+              Array [
+                Object {
+                  "match": Object {
+                    "someFieldThatTellsIfTheResourceIsActive": "AVAILABLE",
+                  },
+                },
+              ],
+            ],
             "must": Array [],
             "must_not": Array [],
           },
@@ -344,7 +384,15 @@ Array [
       "body": Object {
         "query": Object {
           "bool": Object {
-            "filter": Array [],
+            "filter": Array [
+              Array [
+                Object {
+                  "match": Object {
+                    "someFieldThatTellsIfTheResourceIsActive": "AVAILABLE",
+                  },
+                },
+              ],
+            ],
             "must": Array [],
             "must_not": Array [],
           },
@@ -428,7 +476,15 @@ Array [
       "body": Object {
         "query": Object {
           "bool": Object {
-            "filter": Array [],
+            "filter": Array [
+              Array [
+                Object {
+                  "match": Object {
+                    "someFieldThatTellsIfTheResourceIsActive": "AVAILABLE",
+                  },
+                },
+              ],
+            ],
             "must": Array [],
             "must_not": Array [],
           },
@@ -805,7 +861,15 @@ Array [
       "body": Object {
         "query": Object {
           "bool": Object {
-            "filter": Array [],
+            "filter": Array [
+              Array [
+                Object {
+                  "match": Object {
+                    "someFieldThatTellsIfTheResourceIsActive": "AVAILABLE",
+                  },
+                },
+              ],
+            ],
             "must": Array [],
             "must_not": Array [],
           },
@@ -860,7 +924,15 @@ Array [
       "body": Object {
         "query": Object {
           "bool": Object {
-            "filter": Array [],
+            "filter": Array [
+              Array [
+                Object {
+                  "match": Object {
+                    "someFieldThatTellsIfTheResourceIsActive": "AVAILABLE",
+                  },
+                },
+              ],
+            ],
             "must": Array [],
             "must_not": Array [],
           },
@@ -883,7 +955,15 @@ Array [
       "body": Object {
         "query": Object {
           "bool": Object {
-            "filter": Array [],
+            "filter": Array [
+              Array [
+                Object {
+                  "match": Object {
+                    "someFieldThatTellsIfTheResourceIsActive": "AVAILABLE",
+                  },
+                },
+              ],
+            ],
             "must": Array [],
             "must_not": Array [],
           },
@@ -938,7 +1018,15 @@ Array [
       "body": Object {
         "query": Object {
           "bool": Object {
-            "filter": Array [],
+            "filter": Array [
+              Array [
+                Object {
+                  "match": Object {
+                    "someFieldThatTellsIfTheResourceIsActive": "AVAILABLE",
+                  },
+                },
+              ],
+            ],
             "must": Array [],
             "must_not": Array [],
           },
@@ -993,7 +1081,15 @@ Array [
       "body": Object {
         "query": Object {
           "bool": Object {
-            "filter": Array [],
+            "filter": Array [
+              Array [
+                Object {
+                  "match": Object {
+                    "someFieldThatTellsIfTheResourceIsActive": "AVAILABLE",
+                  },
+                },
+              ],
+            ],
             "must": Array [],
             "must_not": Array [],
           },
@@ -1071,7 +1167,15 @@ Array [
       "body": Object {
         "query": Object {
           "bool": Object {
-            "filter": Array [],
+            "filter": Array [
+              Array [
+                Object {
+                  "match": Object {
+                    "someFieldThatTellsIfTheResourceIsActive": "AVAILABLE",
+                  },
+                },
+              ],
+            ],
             "must": Array [],
             "must_not": Array [],
           },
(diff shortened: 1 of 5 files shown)


src/constants.ts
```diff
@@ -10,4 +10,54 @@ export const enum SEARCH_PAGINATION_PARAMS {
     COUNT = '_count',
 }
 
+// Add here which requires date based search
+export const DATEFIELDS = [
```
nguyen102: Nitpick: Can we make these constants `SNAKE_CASE`? Example, `DATE_FIELDS`

src/constants.ts
```diff
@@ -10,4 +10,54 @@ export const enum SEARCH_PAGINATION_PARAMS {
...
+    '_include',
+    '_revinclude',
+    ...ITERATIVE_INCLUSION_PARAMETERS,
+];
+
+export const ALLOWED_PREFIXES = ['eq', 'ne', 'gt', 'lt', 'ge', 'le', 'sa', 'eb', 'ap'];
+
+export const enum PREFIXES {
+    EQUAL = 'eq',
+    NOTEQUAL = 'ne',
```
nguyen102: Nitpick: `NOT_EQUAL`

src/constants.ts
```diff
@@ -10,4 +10,54 @@ export const enum SEARCH_PAGINATION_PARAMS {
...
+    LESSER = 'lt',
+    GREATEROREQUAL = 'ge',
+    LESSEROREQUAL = 'le',
+    STARTSAFTER = 'sa', // These values need to be set appropriately
+    ENDSBEFORE = 'eb', // These values need to be set appropriately
+    APPROXIMATION = 'ap', // These values need to be set appropriately
+}
+
+export const enum ESOPERATORS {
+    LESSEROREQUAL = 'lte',
```
nguyen102: Nitpick: `LESS_OR_EQUAL`

src/elasticSearchService.ts
```diff
@@ -75,43 +66,17 @@ export class ElasticSearchService implements Search {
...
-                        bool: {
-                            must,
-                            filter,
-                        },
-                    },
-                },
+                body,
             };
+
+            console.log(JSON.stringify(params));
```
nguyen102: Debugging statement? Should we delete this?

src/__snapshots__/elasticSearchService.test.ts.snap
```diff
@@ -110,14 +110,9 @@ Array [
       "body": Object {
         "query": Object {
           "bool": Object {
-            "filter": Array [
-              Object {
-                "match": Object {
-                  "someFieldThatTellsIfTheResourceIsActive": "AVAILABLE",
-                },
-              },
-            ],
```
carvantes: This is a bug. Queries must filter using the `filterRulesForActiveResources` clause

btw, if you are not familiar with snapshot tests, the idea is that snapshots should not change unless we are explicitly changing the functionality. I'd expect your changes on this PR should to have none or minimal changes on the existing snapshots since the way we handle dates shouldn't affect the existing queries for non-date fields

https://jestjs.io/docs/en/snapshot-testing

src/queryBuilder.ts
```diff
@@ -0,0 +1,142 @@
...
+        }
+        if (prefix === PREFIXES.ENDSBEFORE) {
+            return ESOPERATORS.LESSEROREQUAL; // need to implement approximation
+        }
+        return prefix;
+    }
+
+    private splitPrefixAndValue(value: string): prefixBasedSearch {
+        // eslint-disable-next-line no-restricted-syntax
+        for (const prefix of ALLOWED_PREFIXES) {
```
carvantes: Using `for of` is discouraged. Use `ALLOWED_PREFIXES.forEach` instead

src/queryBuilder.ts
```diff
@@ -0,0 +1,142 @@
...
+            if (value.includes(prefix)) {
```
carvantes: looks like a bug, what if the value includes prefix but not at the starting position? i.e. "some**eq**stringvalue"

src/queryBuilder.ts
```diff
@@ -0,0 +1,142 @@
+/* eslint-disable class-methods-use-this */
```
carvantes: Let's not do this.

As it stands now, I don't think the `QueryBuilder` should be a class. It has a single public method and its state is not complex and it's just acting as a parameter to the single public method.

I think that this module could simply export a function like:
```
buildQuery(queryParams: any): any
```

carvantes (changes requested):
Hi, 

Thanks for submitting the PR.

Aside from the comments I left in the code, you should also add unit tests for the new functionality. The easiest would be to add tests on `elasticSearchService.test.ts` that cover search queries with dates/prefixes. Tests would be similar to the existing ones and rely on snapshots.

src/constants.ts
PrimKruskal: Thank u nguyen. I have made it as snake case.

src/constants.ts
PrimKruskal: Thank u nguyen. I have made it as snake case.

src/constants.ts
PrimKruskal: Thank u nguyen. I have made it as snake case.

src/elasticSearchService.ts
PrimKruskal: Thank u nguyen. I have removed it.

src/queryBuilder.ts
PrimKruskal: Thank u Carvantes. I have converted to FOR loop, so that we can break once first prefix is found.

src/queryBuilder.ts
PrimKruskal: Thank u Carvantes. As per FHIR standard, date fields will always have prefixes at starting. Now I have handled in a way to ignore if we found prefix other than starting position and considered entire value to part of search string. 
In future while implementing for string,  there won't be search based on value prefix. For string search, we will be having only parameter prefixes like name:contains=eve or name:exact=Eve. 

Kindly correct me, if my understanding is incorrect.

src/queryBuilder.ts
PrimKruskal: Thanks Carvantes. Converted into a function.

src/__snapshots__/elasticSearchService.test.ts.snap
PrimKruskal: Thanks Carvantes.  I have added snapshots but I am not sure whether I have completely addressed your comments.  Kindly check the latest commit and let me know if I need to make any other changes.

src/elasticSearchService.test.ts
```diff
@@ -380,4 +380,51 @@ describe('typeSearch', () => {
...
+            (ElasticSearch.search as jest.Mock).mockResolvedValue(fakeSearchResult);
+            const es = new ElasticSearchService(FILTER_RULES_FOR_ACTIVE_RESOURCES);
+            await es.typeSearch({
+                resourceType: 'Patient',
+                baseUrl: 'https://base-url.com',
+                queryParams,
+                allowedResourceTypes: ALLOWED_RESOURCE_TYPES,
+            });
+            expect((ElasticSearch.search as jest.Mock).mock.calls).toMatchSnapshot('search queries');
+            expect((ElasticSearch.search as jest.Mock).mock.calls).toMatchSnapshot('msearch queries');
```
carvantes: Remove this line 427. Only the _include/_revinclude use msearch queries

carvantes (changes requested):

src/__snapshots__/elasticSearchService.test.ts.snap
carvantes: The latest commit still has the same bug. The snapshot is telling us that a filter clause was removed from existing queries, and that's not what we want.

Queries are missing the `filterRulesForActiveResources` at https://github.com/awslabs/fhir-works-on-aws-search-es/blob/mainline/src/elasticSearchService.ts#L39 clause.

`filterRulesForActiveResources` is a filter that must be present on ALL queries in addition to any clauses that are derived from the search parameters. It is a user provided filter to filter out non-searchable documents. It is useful for soft-deleting documents. 
On the actual implementation the filter is:`[{ match: { documentStatus: 'AVAILABLE' } }]` . On the unit tests the filter is `[{ match: { someFieldThatTellsIfTheResourceIsActive: 'AVAILABLE' } }],`

It shouldn't be too hard to fix, just make sure that your `buildQuery` method always adds a filter clause with the value of `filterRulesForActiveResources`

src/queryBuilder.ts
```diff
@@ -0,0 +1,146 @@
...
+    if (search.prefix === PREFIXES.NOT_EQUAL || search.prefix === PREFIXES.EQUAL) {
+        query = {
+            range: {
+                [searchParameter]: {
+                    gte: search.value,
+                    lte: search.value,
+                },
+            },
+        };
+        return { typeOfQuery: TYPE_OF_QUERY.MUST_NOT, query };
```
carvantes: I believe this is a bug. Why does `PREFIXES.EQUAL` ends up on the `must_not` clause? I believe that no prefix is equivalent to the `eq` prefix

i.e. both are the same:
```
{{API_URL}}/Patient?birthdate=2003-12-03
{{API_URL}}/Patient?birthdate=eq2003-12-03
```

carvantes (changes requested):

src/elasticSearchService.test.ts
PrimKruskal: Thanks Carvantes. Removed search now.

src/queryBuilder.ts
PrimKruskal: Guess morning coffee didn't kick when I coded that line.  Fixed the issue now. Thanks

src/__snapshots__/elasticSearchService.test.ts.snap
PrimKruskal: My bad. I should have captured earlier. I fixed this now. I added additional check to check for length of Filter.  If we have filter available and es is not allowing to add an empty array.  I tested multiple scenarios as well.

src/queryBuilder.ts
```diff
@@ -0,0 +1,152 @@
...
+ * [{ match: { documentStatus: 'AVAILABLE' }}]
+ * @param Returns built ES query
+ */
+
+export function buildQuery(queryParams: any, filterRulesForActiveResources: any): any {
+    const must: any[] = [];
+    const filter: any[] = [];
+    const mustNot: any[] = [];
+    // Filter based on the user request
+    if (filterRulesForActiveResources.length > 0) filter.push(filterRulesForActiveResources);
```
carvantes: There's still an unintended change on the snapshots. the `filter` is now an array of arrays of query objects instead of an array of  query objects. The query itself is valid, but it's better to avoid an unnecessary nested array.

I believe that the following change + regenerating the snapshots should fix it:

```suggestion
    if (filterRulesForActiveResources.length > 0) filter.push(...filterRulesForActiveResources);
```

carvantes (changes requested):

src/__snapshots__/elasticSearchService.test.ts.snap
```diff
@@ -111,13 +111,16 @@ Array [
...
-                  "someFieldThatTellsIfTheResourceIsActive": "AVAILABLE",
+              Array [
+                Object {
+                  "match": Object {
+                    "someFieldThatTellsIfTheResourceIsActive": "AVAILABLE",
+                  },
                 },
-              },
+              ],
             ],
```
carvantes: This should not change to a nested array. See my other comment on the `buildQuery` function

src/queryBuilder.ts
PrimKruskal: Thanks Carvantes. Fixed the issue. Happy Christmas !!!

src/__snapshots__/elasticSearchService.test.ts.snap
PrimKruskal: Thanks Carvantes. Fixed the issue. Happy Christmas !!!

PrimKruskal:
Deleting the pull request.

PR #29, opened 2020-12-14
feat: Add FhirPath parser and compiled search params

carvantes:
Description of changes:

Add a compiler that compiles the base FHIR SearchParameter definitions into a data structure that is useful for building ES queries.

There are two main components.

1. The reduced FHIRPath grammar definition. This grammar is just enough to parse all of the SearchParameter FHIRPath expressions.
1. The compiler script. It follows the same pattern as the revinclude script. It's meant to run locally and have the resulting JSON added to `/src/schema`. The compiler itself is not part of the npm package.

Next step (on a separate PR) is to use the `compiledSearchParameters` files to assemble ES queries.

For example, the FHIRPath expression:
```
OrganizationAffiliation.telecom.where(system='email')
```

 is compiled to:
```js
{
  "resourceType": "OrganizationAffiliation",
  "path": "telecom",
  "condition": ["telecom.system", "=", "email"]
}
```

See https://nearley.js.org/docs/grammar for details about the *.ne file syntax.

Note: It's worth taking a look at the `compiledSearchParameters.4.0.1.json` file during the PR review, even if it is too big for the github diff to display it.

Note: Skipped unit tests on purpose. The compiler already compiled ALL the FHIR v3 & v4 search params. In a sense the resulting JSON files act as snapshot tests for the compiler.


scripts/searchParamsCompiler/run.js
```diff
@@ -0,0 +1,110 @@
...
+the search-parameters.json file into this directory.
+
+You can download the latest FHIR definition from https://www.hl7.org/fhir/downloads.html or find older FHIR versions at http://hl7.org/fhir/directory.html
+
+The nearley compiler needs to be installed separately:
+> npm install -g nearley
+
+First compile the grammar and then run the script:
+> nearleyc reducedFHIRPath.ne -o reducedFHIRPath.js
+> node run.js <fhirVersion>
```
nguyen102, line 17: Should these instructions be in the `README`? Would a user know to use this script without prior instructions to do so? Maybe something along the line of: We have already precompiled FHIR 4.0.1 FHIR definitions. If you would like to support other FHIR definitions please download the new FHIR definition and run the script `scripts/searchParamsCompiler/run.js`.

nguyen102 (commented):
If someone makes a change to `reducedFHIRPath.ne`, what is the procedure for testing that change and ensuring that the grammar still produce the expected `compiledSearchParams`? Would it be worth creating a small subset of the FHIR Search Parameters, saving that to version control as the input and the `compiledSearchParams` as the output? Then we can add a test to run the compiler and compare that to the saved `compiledSearchParams`. 

A smaller dataset would make it easier to understand the expected output given an input.

scripts/searchParamsCompiler/run.js
carvantes: I added unit tests for the complier with hand picked search parameters that cover a variety of scenarios. I agree that this will make it easier to understand and to maintain.

I think we had a similar discussion about docs for revinclude: https://github.com/awslabs/fhir-works-on-aws-search-es/pull/12#discussion_r491116189. In short, I'd rather not update the README since the compiler itself is not part of the npm package. Also, It'd be rare for customers to run this script since it only makes sense to run it for new/different FHIR versions. 

In the future the compiler will be exposed by the search package so that customers can use it to compile IGs. At that point the thorough documentation will be added. Right now the focus is just to generate the JSON files for FHIR 4.0.1 and 3.0.1 for internal use.

nguyen102 (approved):

PR #30, opened 2020-12-15
feat: Updating ElasticSearchService to use new SearchFilter interface

emilhdiaz:
Description of changes:

### DIFF
--- a/src/__snapshots__/elasticSearchService.test.ts.snap
+++ b/src/__snapshots__/elasticSearchService.test.ts.snap
@@ -1263,6 +1263,338 @@ Object {
 }
 `;
 
+exports[`typeSearch filter snapshots for simple filters - not equal 1`] = `
+Array [
+  Array [
+    Object {
+      "body": Object {
+        "query": Object {
+          "bool": Object {
+            "filter": Array [
+              Object {
+                "bool": Object {
+                  "must_not": Array [
+                    Object {
+                      "term": Object {
+                        "someFieldThatTellsIfTheResourceIsActive": "AVAILABLE",
+                      },
+                    },
+                  ],
+                },
+              },
+            ],
+            "must": Array [
+              Object {
+                "query_string": Object {
+                  "default_operator": "AND",
+                  "fields": Array [
+                    "0.*",
+                  ],
+                  "lenient": true,
+                  "query": Object {},
+                },
+              },
+            ],
+          },
+        },
+      },
+      "from": 0,
+      "index": "patient",
+      "size": 20,
+    },
+  ],
+]
+`;
+
+exports[`typeSearch filter snapshots for simple filters - equal 1`] = `
+Array [
+  Array [
+    Object {
+      "body": Object {
+        "query": Object {
+          "bool": Object {
+            "filter": Array [
+              Object {
+                "term": Object {
+                  "someFieldThatTellsIfTheResourceIsActive": "AVAILABLE",
+                },
+              },
+            ],
+            "must": Array [
+              Object {
+                "query_string": Object {
+                  "default_operator": "AND",
+                  "fields": Array [
+                    "0.*",
+                  ],
+                  "lenient": true,
+                  "query": Object {},
+                },
+              },
+            ],
+          },
+        },
+      },
+      "from": 0,
+      "index": "patient",
+      "size": 20,
+    },
+  ],
+]
+`;
+
+exports[`typeSearch filter snapshots for simple filters - greater than 1`] = `
+Array [
+  Array [
+    Object {
+      "body": Object {
+        "query": Object {
+          "bool": Object {
+            "filter": Array [
+              Object {
+                "range": Object {
+                  "age": Object {
+                    "gt": "21",
+                  },
+                },
+              },
+            ],
+            "must": Array [
+              Object {
+                "query_string": Object {
+                  "default_operator": "AND",
+                  "fields": Array [
+                    "0.*",
+                  ],
+                  "lenient": true,
+                  "query": Object {},
+                },
+              },
+            ],
+          },
+        },
+      },
+      "from": 0,
+      "index": "patient",
+      "size": 20,
+    },
+  ],
+]
+`;
+
+exports[`typeSearch filter snapshots for simple filters - less than 1`] = `
+Array [
+  Array [
+    Object {
+      "body": Object {
+        "query": Object {
+          "bool": Object {
+            "filter": Array [
+              Object {
+                "range": Object {
+                  "age": Object {
+                    "lt": "21",
+                  },
+                },
+              },
+            ],
+            "must": Array [
+              Object {
+                "query_string": Object {
+                  "default_operator": "AND",
+                  "fields": Array [
+                    "0.*",
+                  ],
+                  "lenient": true,
+                  "query": Object {},
+                },
+              },
+            ],
+          },
+        },
+      },
+      "from": 0,
+      "index": "patient",
+      "size": 20,
+    },
+  ],
+]
+`;
+
+exports[`typeSearch filter snapshots for simple filters - greater than or equal 1`] = `
+Array [
+  Array [
+    Object {
+      "body": Object {
+        "query": Object {
+          "bool": Object {
+            "filter": Array [
+              Object {
+                "range": Object {
+                  "age": Object {
+                    "gte": "21",
+                  },
+                },
+              },
+            ],
+            "must": Array [
+              Object {
+                "query_string": Object {
+                  "default_operator": "AND",
+                  "fields": Array [
+                    "0.*",
+                  ],
+                  "lenient": true,
+                  "query": Object {},
+                },
+              },
+            ],
+          },
+        },
+      },
+      "from": 0,
+      "index": "patient",
+      "size": 20,
+    },
+  ],
+]
+`;
+
+exports[`typeSearch filter snapshots for simple filters - less than or equal 1`] = `
+Array [
+  Array [
+    Object {
+      "body": Object {
+        "query": Object {
+          "bool": Object {
+            "filter": Array [
+              Object {
+                "range": Object {
+                  "age": Object {
+                    "lte": "21",
+                  },
+                },
+              },
+            ],
+            "must": Array [
+              Object {
+                "query_string": Object {
+                  "default_operator": "AND",
+                  "fields": Array [
+                    "0.*",
+                  ],
+                  "lenient": true,
+                  "query": Object {},
+                },
+              },
+            ],
+          },
+        },
+      },
+      "from": 0,
+      "index": "patient",
+      "size": 20,
+    },
+  ],
+]
+`;
+
+exports[`typeSearch filter snapshots for simple filters - AND combination 1`] = `
+Array [
+  Array [
+    Object {
+      "body": Object {
+        "query": Object {
+          "bool": Object {
+            "filter": Array [
+              Object {
+                "bool": Object {
+                  "filter": Array [
+                    Object {
+                      "term": Object {
+                        "someFieldThatTellsIfTheResourceIsActive": "AVAILABLE",
+                      },
+                    },
+                    Object {
+                      "term": Object {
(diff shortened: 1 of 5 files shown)


src/elasticSearchService.ts
```diff
@@ -66,7 +68,7 @@ export class ElasticSearchService implements Search {
     searchParams => {field: value}
      */
     async typeSearch(request: TypeSearchRequest): Promise<SearchResponse> {
-        const { queryParams, resourceType } = request;
+        const { queryParams, searchFilters, resourceType } = request;
```
emilhdiaz, line 70: Note: Do we want to completely replace passing in `queryParams` in favor of expressing those as `searchFilters` by the calling code, or continue to do the conversion into `SearchFilters` as currently done below?

src/elasticSearchService.ts
```diff
@@ -36,28 +38,28 @@ const MAX_INCLUDE_ITERATIVE_DEPTH = 5;
...
+     * @param searchFiltersForAllQueries - If you are storing both History and Search resources
+     * in your elastic search you can filter out your History elements by supplying a list of SearchFilters
+     *
      * @param cleanUpFunction - If you are storing non-fhir related parameters pass this function to clean
      * the return ES objects
      * @param fhirVersion
      */
     constructor(
-        filterRulesForActiveResources: any[] = [],
+        searchFiltersForAllQueries: SearchFilter[] = [],
```
rsmayda, line 55: +1 nice

src/elasticSearchService.ts
rsmayda: So technically queryParams can be fuzzy searching filtering is a binary yes or no

There is a use case for both

rsmayda: Here is a good overview of must vs filter https://www.elastic.co/guide/en/elasticsearch/reference/current/query-filter-context.html

rsmayda (commented):
Have you had to test this? This one will be challenging maybe querying ES directly playing with the different filtering options

src/elasticSearchService.ts
```diff
@@ -78,26 +80,17 @@ export class ElasticSearchService implements Search {
...
             const searchParameterToValue = { ...queryParams };
+            console.log('searchParameterToValue:', searchParameterToValue);
+            const must: SearchFilter[] = ElasticSearchService.searchFiltersToElasticQuery(
+                this.queryParamsToSearchFilter(searchParameterToValue),
+            );
+            console.log('must:', must);
 
-            const must: any = [];
-            // TODO Implement fuzzy matches
-            Object.entries(searchParameterToValue).forEach(([searchParameter, value]) => {
```
rsmayda: hmmm for code change perspective lets focus on just changing the filter behavior and not modify the existing `must`

src/elasticSearchService.ts
```diff
@@ -333,4 +326,97 @@ export class ElasticSearchService implements Search {
...
+                    filterOperator: '~',
+                };
+            });
+    }
+
+    private static searchFilterToElasticQuery(searchFilter: SearchFilter): any {
+        const { filterKey, filterValue, filterOperator } = searchFilter;
+
+        switch (filterOperator) {
+            case '~': {
```
rsmayda: From a filter perspective I don't know if we should have approximate as the filter is supposed to be a simple yes/no approximate is a bit wishy washy

src/elasticSearchService.ts
emilhdiaz: Hmm, I see your point. I noticed however that in the unit tests we expect `match` queries (which analyzes the query values and adds some level of fuzziness) instead of `term` queries (which are exact) in the filter context. How binary should filtering be? Should we convert over to using `term` queries instead? 
@nguyen102 FYI

emilhdiaz:
> Have you had to test this? This one will be challenging maybe querying ES directly playing with the different filtering options

I did for the basic `==` scenario, haven't yet for the other scenarios. Working through a local ES setup to make this easier to test.

src/elasticSearchService.ts
```diff
@@ -97,7 +98,11 @@ export class ElasticSearchService implements Search {
                 must.push(query);
             });
 
-            const filter = this.filterRulesForActiveResources;
+            const filter: SearchFilter[] = ElasticSearchService.searchFiltersToElasticQuery([
+                ...this.searchFiltersForAllQueries,
+                ...(searchFilters ?? []),
+            ]);
+            console.log('filter:', filter);
```
emilhdiaz: Note to self to remove these log statements...

src/elasticSearchService.ts
emilhdiaz: Good point, I reverted these changes.

src/elasticSearchService.ts
emilhdiaz: Yep, now that I understand the difference in the exactness of filters vs matches, this makes sense. I dropped this extra operator.

emilhdiaz: This approximate operator has been removed

src/elasticSearchService.ts
```diff
@@ -247,15 +252,15 @@ export class ElasticSearchService implements Search {
         const includeSearchQueries = buildIncludeQueries(
             request.queryParams,
             searchEntries.map(x => x.resource),
-            this.filterRulesForActiveResources,
+            ElasticSearchService.searchFiltersToElasticQuery(this.searchFiltersForAllQueries),
```
rsmayda: Hmm I wonder if we should add the other filters to revInclude/include searches. Potential change would make this just be `filter` const that you created above

@carvantes thoughts?

src/elasticSearchService.ts
```diff
@@ -247,15 +252,15 @@ export class ElasticSearchService implements Search {
...
             this.fhirVersion,
             iterative,
         );
 
         const revIncludeSearchQueries = buildRevIncludeQueries(
             request.queryParams,
             searchEntries.map(x => x.resource),
-            this.filterRulesForActiveResources,
+            ElasticSearchService.searchFiltersToElasticQuery(this.searchFiltersForAllQueries),
```
rsmayda: Same as 255

src/elasticSearchService.ts
```diff
@@ -333,4 +338,96 @@ export class ElasticSearchService implements Search {
...
+            case '==': {
+                return {
+                    term: {
+                        [key]: value,
+                    },
+                };
+            }
+            case '!=': {
+                return {
+                    bool: {
```
rsmayda, line 362: Do we want this second `bool`? I think the result of `searchFilterToElasticQuery` with a != & OR will result in:
```json
{
  "query":{
    "bool":{
      "filter":[
        {
          "bool":{
            "bool":{
              "must_not":[
                {
                  "term":{
                    "..."
                  }
                }
              ]
            }
          }
        }
      ]
    }
  }
}
```

src/elasticSearchService.ts
carvantes: I think that they should be added as well. If `searchFilters` is being used to enforce SMART scopes, then those filters MUST be part of revinclude/include queries as well

src/elasticSearchService.ts
emilhdiaz: @rsmayda / @nguyen102 I tested converting the `match` filters to `term` filters and the system seems to behave normally. Did you come to a conclusion on whether this change should be part of this MR or a separate MR?

nguyen102: @emilhdiaz after talking with the team, let's stick with `match` for now, and keep this PR focused on just the filter functionality. Currently we don't have the bandwidth to test the `term` filters before the holiday break.

src/elasticSearchService.ts
emilhdiaz: Good catch, I updated to make sure that revInclude/include searches also honor the `SearchFilter`'s from `TypeSearchRequest`

src/elasticSearchService.ts
emilhdiaz: Updated

src/elasticSearchService.ts
emilhdiaz: Currently you end up with this: 

```
{
  "query":{
    "bool":{
      "filter":[
        {
          "bool":{
            "should": [
              {
                "bool": {
                  "must_not":[
                    {
                      "term":{
                        "..."
                      }
                    }
                  ]
                }
              },
              {
                "bool": {
                  "must_not":[
                    {
                      "term":{
                        "..."
                      }
                    }
                  ]
                }
              }
            ]
          }
        }
      ]
    }
  }
}
```

src/elasticSearchService.ts
```diff
@@ -244,29 +248,35 @@ export class ElasticSearchService implements Search {
         request: TypeSearchRequest,
         iterative?: true,
     ): Promise<SearchEntry[]> {
+        const { queryParams, searchFilters, allowedResourceTypes, baseUrl } = request;
+        const filter: SearchFilter[] = ElasticSearchService.searchFiltersToElasticQuery([
+            ...this.searchFiltersForAllQueries,
+            ...(searchFilters ?? []),
+        ]);
```
rsmayda: Could this be passed in as a param?

src/elasticSearchService.ts
emilhdiaz: Thanks @nguyen102, I reverted that commit to preserve the previous behavior.

src/elasticSearchService.ts
emilhdiaz: It could, but the code already passes and unpacks `TypeSearchRequest`, if we passed in `filters` in addition to `TypeSearchRequest`, you'd end up with 2 sources of truth for filter. Thoughts?

src/__snapshots__/elasticSearchService.test.ts.snap
```diff
@@ -1263,6 +1263,338 @@ Object {
...
+                },
+              },
+            ],
+            "must": Array [
+              Object {
+                "query_string": Object {
+                  "default_operator": "AND",
+                  "fields": Array [
+                    "0.*",
+                  ],
```
carvantes: looks like a bug

src/elasticSearchService.test.ts
```diff
@@ -380,4 +389,132 @@ describe('typeSearch', () => {
...
+                        ],
+                    },
+                },
+            };
+            (ElasticSearch.search as jest.Mock).mockResolvedValue(fakeSearchResult);
+            const es = new ElasticSearchService(searchFilters);
+            await es.typeSearch({
+                resourceType: 'Patient',
+                baseUrl: 'https://base-url.com',
+                queryParams: [{}],
```
carvantes:
```suggestion
                queryParams: {},
```
queryParams is an object.  That should get rid of the `0.*` in the snapshots

src/elasticSearchService.ts
```diff
@@ -333,4 +343,96 @@ export class ElasticSearchService implements Search {
...
+                    },
+                };
+            }
+            default: {
+                throw new Error('Unknown comparison operator');
+            }
+        }
+    }
+
+    private static searchFilterToElasticQuery(searchFilter: SearchFilter): any {
```
nguyen102: To keep naming consistent, I would rename this to `buildElasticQuery`.  The naming format `aToB` is used in our code as a way to name a map. For example `idToPatient` would be a map with the key of `id` and a value of `map`.

src/elasticSearchService.test.ts
emilhdiaz: Ah, yep. Copy/paste error on my part. Fixed.

src/__snapshots__/elasticSearchService.test.ts.snap
emilhdiaz: Thanks, fixed!

src/elasticSearchService.ts
emilhdiaz: Agreed, thanks for the feedback. Updated to `buildElasticSearchFilter`

nguyen102 (approved):
Looks good! Thanks for your hard work on this Emil!

PR #31, opened 2020-12-17
feat: use compiled search params to build ES queries

carvantes:
Description of changes:

Use the compiled search parameters to build ES queries. 

The basic idea is to:
1. lookup the compiled search parameter
1.2 Throw an error if the param is invalid
2. build the ES query using the path from the compiled FHIR search parameter (instead of the raw parameter as before) 
3. Optionally, add additional query clauses if the compiled FHIR search parameter has conditions

Note: The current implementation does not support the specifics for the different search param types (date, token, etc.). It just treats everything as string. This PR does not change that.

### DIFF
--- a/src/FHIRSearchParametersRegistry/index.ts
+++ b/src/FHIRSearchParametersRegistry/index.ts
@@ -1,0 +1,45 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+import { FhirVersion } from 'fhir-works-on-aws-interface';
+import compiledSearchParamsV4 from '../schema/compiledSearchParameters.4.0.1.json';
+import compiledSearchParamsV3 from '../schema/compiledSearchParameters.3.0.1.json';
+
+export type SearchParam = {
+    type: string;
+    description: string;
+    target?: string[];
+    compiled: { resourceType: string; path: string; condition?: string[] }[];
+};
+export type CompiledSearchParams = {
+    [resourceType: string]: {
+        [name: string]: SearchParam;
+    };
+};
+
+/**
+ * This class is the single authority over the supported FHIR SearchParameters and their definitions
+ */
+// eslint-disable-next-line import/prefer-default-export
+export class FHIRSearchParametersRegistry {
+    private readonly compiledSearchParams: CompiledSearchParams;
+
+    constructor(fhirVersion: FhirVersion) {
+        if (fhirVersion === '4.0.1') {
+            this.compiledSearchParams = compiledSearchParamsV4 as any;
+        } else {
+            this.compiledSearchParams = compiledSearchParamsV3 as any;
+        }
+    }
+
+    /**
+     * Retrieve a search parameter. Returns undefined if the parameter is not found on the registry.
+     * @param resourceType FHIR resource type
+     * @param name search parameter name
+     */
+    getSearchParameter(resourceType: string, name: string): SearchParam | undefined {
+        return this.compiledSearchParams?.[resourceType]?.[name];
+    }
+}
--- a/src/FHIRSearchParametersRegistry/index.ts
+++ b/src/FHIRSearchParametersRegistry/index.ts
@@ -7,8 +7,43 @@ import { FhirVersion } from 'fhir-works-on-aws-interface';
 import compiledSearchParamsV4 from '../schema/compiledSearchParameters.4.0.1.json';
 import compiledSearchParamsV3 from '../schema/compiledSearchParameters.3.0.1.json';
 
+/**
+ * type: SearchParameter type
+ *
+ * description: SearchParameter description
+ *
+ * target: SearchParameter target. Only used for parameters with type reference
+ *
+ * compiled: array of objects that can be used to build ES queries. In most cases there is a single element in the array. Multiple elements in the array have on OR relationship between them
+ *
+ * compiled[].resourceType: FHIR resource type
+ *
+ * compiled[].path: object path to be used as field in queries
+ *
+ * compiled[].condition: a 3 element array with the elements of a condition [field, operator, value]
+ *
+ * @example
+ * {
+ *    "type": "reference",
+ *    "description": "What resource is being referenced",
+ *    "target": [
+ *      "Library",
+ *      "Account",
+ *      "ActivityDefinition",
+ *    ],
+ *    "compiled": [
+ *      {
+ *        "resourceType": "ActivityDefinition",
+ *        "path": "relatedArtifact.resource",
+ *        "condition": ["relatedArtifact.type", "=", "depends-on"]
+ *      },
+ *      {"resourceType": "ActivityDefinition", "path": "library"}
+ *    ]
+ *  }
+ *
+ */
 export type SearchParam = {
-    type: string;
+    type: 'composite' | 'date' | 'number' | 'quantity' | 'reference' | 'special' | 'string' | 'token' | 'uri';
     description: string;
     target?: string[];
     compiled: { resourceType: string; path: string; condition?: string[] }[];
@@ -39,7 +74,7 @@ export class FHIRSearchParametersRegistry {
      * @param resourceType FHIR resource type
      * @param name search parameter name
      */
-    getSearchParameter(resourceType: string, name: string): SearchParam? {
+    getSearchParameter(resourceType: string, name: string): SearchParam | undefined {
         return this.compiledSearchParams?.[resourceType]?.[name];
     }
 }
--- a/src/FHIRSearchParametersRegistry/index.ts
+++ b/src/FHIRSearchParametersRegistry/index.ts
@@ -39,7 +39,7 @@ export class FHIRSearchParametersRegistry {
      * @param resourceType FHIR resource type
      * @param name search parameter name
      */
-    getSearchParameter(resourceType: string, name: string): SearchParam | undefined {
+    getSearchParameter(resourceType: string, name: string): SearchParam? {
         return this.compiledSearchParams?.[resourceType]?.[name];
     }
 }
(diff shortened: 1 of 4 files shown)


src/elasticSearchService.test.ts
```diff
@@ -33,14 +34,13 @@ describe('typeSearch', () => {
             [{}],
             [{ _count: 10, _getpagesoffset: 2 }],
             [{ gender: 'female', name: 'Emily' }],
-            [{ id: '11111111-1111-1111-1111-111111111111' }],
```
carvantes, line 36: `id` is no longer a valid parameter. `_id` is the correct one

rsmayda: why? Is this true from a FHIR perspective? should we change the persistence layer to save id as _id as an additional field?

src/elasticSearchService.ts
```diff
@@ -81,20 +89,63 @@ export class ElasticSearchService implements Search {
...
-                must.push(query);
+                const value = escapeQueryString(searchValue as string);
+                const fhirSearchParam = this.fhirSearchParams.getSearchParameter(resourceType, searchParameter);
+                if (fhirSearchParam === undefined) {
+                    throw new InvalidSearchParameterError(
+                        `Invalid search parameter '${searchParameter}' for resource type ${resourceType}`,
+                    );
+                }
+
+                const queries = fhirSearchParam.compiled.map(compiled => {
```
rsmayda, line 104: I don't fully understand this loop;;
```
for each search parameters:
   for each compiled parameter:
```

Could you explain a use case for when you have multiple compiled params?

src/elasticSearchService.ts
```diff
@@ -81,20 +89,63 @@ export class ElasticSearchService implements Search {
...
+                            query: value,
+                            default_operator: 'AND',
+                            lenient: true,
+                        },
+                    };
+
+                    // In most cases conditions are used for fields that are an array of objects
+                    // Ideally we should be using a nested query, but that'd require to update the index mappings.
+                    //
+                    // Simply using an array of bool.must is good enough for most cases. The result will contain the correct documents, however it MAY contain additional documents
```
rsmayda, line 119: `however it MAY contain additional documents` -- could you explain this more? What will the additional documents be, if they will all be correct?

src/FHIRSearchParametersRegistry/index.ts
```diff
@@ -0,0 +1,45 @@
...
+            this.compiledSearchParams = compiledSearchParamsV3 as any;
+        }
+    }
+
+    /**
+     * Retrieve a search parameter. Returns undefined if the parameter is not found on the registry.
+     * @param resourceType FHIR resource type
+     * @param name search parameter name
+     */
+    getSearchParameter(resourceType: string, name: string): SearchParam | undefined {
```
rsmayda, line 77: Can you use `?` for returns
```suggestion
    getSearchParameter(resourceType: string, name: string): SearchParam? {
```

src/FHIRSearchParametersRegistry/index.ts
```diff
@@ -0,0 +1,45 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+import { FhirVersion } from 'fhir-works-on-aws-interface';
+import compiledSearchParamsV4 from '../schema/compiledSearchParameters.4.0.1.json';
+import compiledSearchParamsV3 from '../schema/compiledSearchParameters.3.0.1.json';
+
+export type SearchParam = {
```
rsmayda, line 45: Could you add comments explaining the params; especially `compiled.condition`; if you use long winded comments, like :
```suggestion
/**
* <comment here>
*/
export type SearchParam = {
```

My IDE (VSCode) will add those comments when I hover over time object

src/elasticSearchService.test.ts
carvantes: This is correct per the FHIR spec. https://www.hl7.org/fhir/search.html#id

`id` was never a truly valid parameter, it's just that before we used the parameters directly as fields on ES documents.

There's no need to change persistence. The expression for `_id` maps it to the `id` field:

```
...
"name" : "_id",
"base" : ["Resource"],
"expression" : "Resource.id",
...
```

src/elasticSearchService.ts
carvantes: This only occurs when there are OR conditions i.e. https://github.com/awslabs/fhir-works-on-aws-search-es/blob/feat-fhir-search-params/scripts/searchParamsCompiler/__snapshots__/compile.test.js.snap#L138-L158

src/elasticSearchService.ts
carvantes: https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html explains it quite well.

Also, to clarify **not all** the returned documents are correct matches, but the correct documents will always part of the results.

Based on the examples from the ES docs, consider the documents:
```
{
  "group" : "fans",
  "user" : [ 
    {
      "first" : "John",
      "last" :  "Smith"
    },
    {
      "first" : "Alice",
      "last" :  "White"
    }
  ]
}
{
  "group" : "fans",
  "user" : [ 
    {
      "first" : "John",
      "last" :  "xxxx"
    },
    {
      "first" : "yyyy",
      "last" :  "Smith"
    }
  ]
}
```

a query like `user.first="John" and user.last="Smith"` will match **both** documents. Only the 1st one is correct, the 2nd one is an additional match that we were not expecting.

src/elasticSearchService.ts
```diff
@@ -81,20 +89,63 @@ export class ElasticSearchService implements Search {
...
-                        default_operator: 'AND',
-                        lenient: true,
-                    },
-                };
-                must.push(query);
+                const value = escapeQueryString(searchValue as string);
+                const fhirSearchParam = this.fhirSearchParams.getSearchParameter(resourceType, searchParameter);
+                if (fhirSearchParam === undefined) {
+                    throw new InvalidSearchParameterError(
+                        `Invalid search parameter '${searchParameter}' for resource type ${resourceType}`,
```
nguyen102, line 100: Should we just pass back the relevant params to `InvalidSearchParameterError` and then have `errorHandling` in the routing package form the error message?

So in this case `throw new InvalidSearchParameterError({searchParameter, resourceType})`

The thought being, since this is `InvalidSearchParameterError`, repeating `Invalid search parameter` seems repetitive.

src/FHIRSearchParametersRegistry/index.ts
carvantes: added docs

src/FHIRSearchParametersRegistry/index.ts
carvantes: not really. Your suggested change is invalid syntax.

https://github.com/microsoft/TypeScript/issues/13184

src/elasticSearchService.ts
carvantes: We could do that, but I don't see a major benefit in doing so. 

Granted, we don't have a standard place to build the error messages right now. Some msgs are a default on the error object, some msgs are set when throwing an error(like this one), and some are set by the router error handler. 

Also, repetition is sort of unavoidable since we are saying the same thing but in user friendly text. Moving it to the router it will be still be "repeating" the `"Invalid search parameter"` for an `InvalidSearchParameterError `

nguyen102: If we move "Invalid search parameter" to `errorHandling` in the router we won't have to include "Invalid search parameter" when throwing `InvalidSearchParameterError` in other places. We can just pass the relevant `searchParameter` and `resourceType` to `InvalidSearchParameterError`. 

Otherwise, if we're forming the full message each time we throw `InvalidSearchParameterError`, it seems we can just create a more generic error such as `InvalidParameterError`.

carvantes: I guess I like the flexibility of just adding a string error message. Right now there's only a single error, but later on we may add more detailed error messages for things like invalid modifiers, invalid prefixes, malformed tokens, etc. I'd rather not parametrize the error since it is tedious to update interface + search + routing just to add a more detailed error msg. 

Also, ideally search should have a single place where it validates parameters. If this error is being thrown in many different places across the codebase, then maybe there's a more important refactor to do.

nguyen102: That sounds reasonable. Let's keep this flexible for now, and we can come back to this in the future, if we find we're reusing the same error message.

src/elasticSearchService.ts
```diff
@@ -81,20 +89,63 @@ export class ElasticSearchService implements Search {
...
+                    // Simply using an array of bool.must is good enough for most cases. The result will contain the correct documents, however it MAY contain additional documents
+                    // https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html
+                    if (compiled.condition !== undefined) {
+                        return {
+                            bool: {
+                                must: [
+                                    pathQuery,
+                                    {
+                                        query_string: {
+                                            fields: [compiled.condition[0], `${compiled.condition[0]}.*`],
```
nguyen102, line 128: `compiled[].condition: a 3 element array with the elements of a condition [field, operator, value]`

Do we need to take into account the `operator`? What if the operator was `>` or `<`? The current query seems to be for the case in which the operation is `=`.

carvantes: We can assume it is always `=`.  Our grammar only supports the `=` operator
https://github.com/awslabs/fhir-works-on-aws-search-es/blob/feat-fhir-search-params/scripts/searchParamsCompiler/reducedFHIRPath.ne#L22

I haven't found any search parameter that uses `>` or `<`

src/elasticSearchService.ts
rsmayda: Ahh ok thanks

nguyen102 (approved):

rsmayda (approved):

PR #33, opened 2020-12-23
feat: Use compiled search params for _include and _revinclude queries

carvantes:
Description of changes:

Use the compiled search params to build _include and _revinclude queries instead of the old `fhirResourceReferencesMatrix` json files

Notes:
- Updated the compiler to add a few more fields. `base` and `name` are just convenient and the `url` may show up on the capability statement. 
- Functionality remains mostly the same. I updated tests to use the correct search parameter. i.e. `MedicationStatement:partOf -> MedicationStatement:part-of`
- The only functional change is that now we only allow searching on valid search parameters instead of any field of type reference. This also means that wildcards now include less resources. This is correct. 
Not every property of type reference has a corresponding search parameter. i.e. `MedicationRequest.basedOn` is of type reference but there's no search parameter for it, so it is not available for include/revinclude queries.


src/FHIRSearchParametersRegistry/index.ts
```diff
@@ -43,8 +52,11 @@ import compiledSearchParamsV3 from '../schema/compiledSearchParameters.3.0.1.jso
  *
  */
 export type SearchParam = {
+    name: string;
```
rsmayda, line 55: nit: Is there a better `name` here? name seems too vague. Perhaps `searchParamName` or `paramName`

src/FHIRSearchParametersRegistry/index.ts
```diff
@@ -73,8 +85,62 @@ export class FHIRSearchParametersRegistry {
...
+    /**
+     * Retrieve a search parameter of type "reference"
+     * @param resourceType
+     * @param name
+     * @param targetResourceType
+     * @return the matching SearchParam or undefined if there's no match
+     */
+    getReferenceSearchParameter(
+        resourceType: string,
+        name: string,
```
rsmayda, line 131: nit: Is there a better `name` here? name seems too vague. Perhaps `searchParamName` or `paramName`

src/FHIRSearchParametersRegistry/index.ts
```diff
@@ -73,8 +85,62 @@ export class FHIRSearchParametersRegistry {
...
+     */
+    getIncludeSearchParameters(resourceType: string): SearchParam[] {
+        return Object.values(this.compiledSearchParams?.[resourceType]).filter(s => s.type === 'reference');
+    }
+
+    /**
+     * Retrieve all the SearchParams that can be used in _revinclude queries for a given resource type
+     * @param resourceType
+     */
+    getRevIncludeSearchParameters(resourceType: string): SearchParam[] {
```
rsmayda, line 162: instead of computing this on the fly what are your thoughts on pre-computing this as a json/map during the compile step?

src/searchInclusions.ts
```diff
@@ -4,17 +4,16 @@
...
 import { isPresent } from './tsUtils';
+import { FHIRSearchParametersRegistry } from './FHIRSearchParametersRegistry';
 
 export type InclusionSearchParameter = {
     type: '_include' | '_revinclude';
     isWildcard: false;
     isIterate?: true;
     sourceResource: string;
     searchParameter: string;
+    path?: string;
```
rsmayda, line 33: What is an example of the path? when would you use/not use?

src/searchInclusions.ts
```diff
@@ -199,22 +196,32 @@ export const buildRevIncludeQuery = (
...
-        return resourceReferencesMatrixV3;
-    }
-    return [];
+            if (searchParam === undefined) {
+                return undefined;
+            }
+
+            return { ...includeParam, path: searchParam.compiled[0].path };
+        })
+        .filter(isPresent);
```
rsmayda: Could we do the filter first then map? -- I believe that is a minor optimization

src/FHIRSearchParametersRegistry/index.ts
carvantes: I prefer to keep it as is.

I think it's redundant to call it `searchParamName` within a type that is already called `SearchParam`. Also, more importantly, I'm keeping the exact same field names that FHIR uses for `SearchParameters`. FHIR simply calls it `name` on the JSON definitions and on the docs.

src/FHIRSearchParametersRegistry/index.ts
carvantes: same as above

rsmayda: In this case this is just a method and not part of an object -- if I were to call this method and had that `name` parameter, would I know that `name` represents the name of the search parameter in question? happy to flex here as well but I feel this is a different scenario

carvantes: It's a method named `getReferenceSearchParameter` on a class named `FHIRSearchParametersRegistry`. I think that it is clear that `name` refers to a search parameter name in this context (In the same way that I'm not using `searchParamResourceType` instead of simply `resourceType`)

I can rename it if you feel strongly about it.

rsmayda: Fine to leave as is

src/searchInclusions.ts
carvantes: path is the actual object path where the reference value can be found.

i.e. for `ImmunizationRecommendation` the search param `information` has the path `recommendation.supportingPatientInformation`

All valid search params have a path. But it is optional here since `InclusionSearchParameter` is first built from the query params and then the path is added afterwards If it is indeed a valid search parameter.

src/searchInclusions.ts
carvantes: I rewrote this to not use map/filter

src/FHIRSearchParametersRegistry/index.ts
carvantes: yeah, I was thinking of something similar and ended up changing things a bit. 

It was awkward for the compiler to worry about how the parameters were being looked up ( by base & name, finding the ones with type reference, etc.). I changed it to simply return an array of search parameters.

I changed `FHIRSearchParametersRegistry` to loop through the compiled search params **once** on the constructor and store the data structures that are ideal for the lookups that are needed(by base and name, include, revinclude).

src/searchInclusions.ts
rsmayda: Ahh super helpful example if you have a moment could you add in this comment explaining what the path is and how the param name != path

carvantes: I added docs with an example for this type

rsmayda (approved):

PR #35, opened 2021-01-06
feat: implement getCapabilities

carvantes:
Description of changes:
Implement the new `getCapabilities` interface method. The capability statement is built only once and then cached in a class variable.

Related PRs:
- https://github.com/awslabs/fhir-works-on-aws-interface/pull/42


src/FHIRSearchParametersRegistry/index.ts
```diff
@@ -80,17 +87,20 @@ export class FHIRSearchParametersRegistry {
         };
     };
 
+    private readonly capabilityStatement: SearchCapabilityStatement;
+
     constructor(fhirVersion: FhirVersion) {
         let compiledSearchParams: SearchParam[];
         if (fhirVersion === '4.0.1') {
-            compiledSearchParams = compiledSearchParamsV4 as SearchParam[];
+            compiledSearchParams = (compiledSearchParamsV4 as unknown) as SearchParam[];
```
nguyen102: interesting, what does asserting  `compiledSearchParamsV4` twice do?

src/FHIRSearchParametersRegistry/index.ts
```diff
@@ -107,6 +117,32 @@ export class FHIRSearchParametersRegistry {
...
+            this.capabilityStatement[resourceType].searchRevInclude = [
+                '*',
+                ...(this.revincludeMap[resourceType]?.map(x => `${x.base}:${x.name}`) ?? []),
+            ];
+        });
+
+        const resourceSearchParams = Object.values(this.typeNameMap.Resource).map(toCapabilityStatement);
+
+        Object.values(this.capabilityStatement).forEach(x => {
+            x.searchParam.push(...resourceSearchParams);
```
nguyen102: On my IDE it shows a linting error for `x`.
`TS2571: Object is of type 'unknown'.`

carvantes: I'm not sure why. btw, that's not an eslint error, that's a typescript error. Such an error would make typescript compilation fail, but the whole `yarn release` runs fine on my end, so it's likely to be an issue on your machine.

1. Do you also have the updated interface from my other PR locally? that var is expected to be of the newly introduced `SearchCapabilities` type.
1. IDEs sometimes briefly show a var type as unknown when opening/modifying a file(I'm guessing that while waiting for tsc to run in the background). At least webstorm does that, it does not highlight it as an error though.

src/FHIRSearchParametersRegistry/index.ts
carvantes: I reverted this. It's not necessary.

I added it when I was testing some changes on the compiled search params JSON file. Typescript makes checks to see if a type casting is safe, but you can bypass them by casting `T1` into `unknown` into `T2`. It's not safe, but sometimes it's convenient for quick tests.

src/FHIRSearchParametersRegistry/index.ts
nguyen102: Ahh, that makes sense. My IDE is erroring out on the interface issue. It must be because I don't have your interface changes.

nguyen102 (approved):

src/FHIRSearchParametersRegistry/index.ts
```diff
@@ -107,6 +117,33 @@ export class FHIRSearchParametersRegistry {
                 });
             }
         });
+
+        Object.entries(this.typeNameMap).forEach(([resourceType, searchParams]) => {
```
rsmayda, line 121: This code is in the constructor when is the `FHIRSearchParametersRegistry` initialized? I am concerned that this code is ran for use-cases when the request is just `GET /Patient?name=Rob`

carvantes: The time it takes to build the capability statement is negligible ( < 10ms. For such small values measuring time with `console.time` isn't even reliable. doing "absolutely nothing" (i.e.`console.time('x'); console.timeEnd('x')`) also takes ~10ms ). 

It could be refactored to lazily build the cap statement on the 1st call to `getCapabilities`, but It's not worth it and it is simpler to keep it as is.

rsmayda (approved):

PR #36, opened 2021-01-12
feat: support standard FHIR search parameters

carvantes:
Description of changes:

This is the final merge of the feature branch into mainline. The changes have already been on previous PRs: #29 #31 #32 #33 #35 

The only new changes are bumping the interface version and a couple very minor changes when fixing merge conflicts from mainline


scripts/searchParamsCompiler/compile.js
```diff
@@ -0,0 +1,82 @@
...
+ */
+
+/*
+This scripts generates the JSON files at src/schema. Before running the script download the JSON FHIR definition package and copy
+the search-parameters.json file into this directory.
+
+You can download the latest FHIR definition from https://www.hl7.org/fhir/downloads.html or find older FHIR versions at http://hl7.org/fhir/directory.html
+
+The nearley compiler needs to be installed separately:
+> npm install -g nearley
```
rsmayda, line 13: Should we advise yarn here since we use yarn elsewhere?

scripts/searchParamsCompiler/reducedFHIRPath.js
```diff
@@ -0,0 +1,49 @@
+// Generated automatically by nearley, version 2.20.0
```
rsmayda, line 1: Add licensing & a comment say please do not modify this file manually

scripts/searchParamsCompiler/run.js
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+/*
+This scripts generates the JSON files at src/schema. Before running the script download the JSON FHIR definition package and copy
+the search-parameters.json file into this directory.
+
+You can download the latest FHIR definition from https://www.hl7.org/fhir/downloads.html or find older FHIR versions at http://hl7.org/fhir/directory.html
+
+If you are modifying the grammar at reducedFHIRPath.ne you need to compile it. The nearley compiler needs to be installed separately:
+> npm install -g nearley
```
rsmayda, line 13: Advise on yarn usage here instead?

src/schema/compiledSearchParameters.3.0.1.json
```diff
@@ -0,0 +1,18343 @@
+[
```
rsmayda: Minify both jsons for a smaller file size (i.e. remove all spacing & new lines)

rsmayda (changes requested):
Add license and minify json

scripts/searchParamsCompiler/compile.js
carvantes: Not really. This is a global install so there's no real need to use yarn. We could use yarn here but it is far more common to do global installs with npm.

similar to https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/scripts/install.sh#L70

scripts/searchParamsCompiler/run.js
carvantes: same as above

scripts/searchParamsCompiler/reducedFHIRPath.js
carvantes: AFAIK there's no need to add license headers to generated code. Otherwise you'd need to manually reapply the license headers everytime you regenerate the file. The header is mandatory on the source code that generates the generated code though(in this case the *.ne file). 

Also the `please do not modify this file manually` is kind of implied by the `Generated automatically by nearley` and adding it creates the same issue that it'll be gone when the file is regenerated.

I think that this file can be kept as is.

src/schema/compiledSearchParameters.3.0.1.json
carvantes: done

rsmayda (approved):

PR #38, opened 2021-01-20
feat: Add ImplementationGuides compile method

carvantes:
Description of changes:

Mostly just refactoring the old `.js` compiler script into `.ts` to use it as the implementation of the `ImplementationGuides` interface `compile` method. The functionality of the compiler remains unchanged.

We are still bundling the FHIR R4 resources into the search package, so we retain `scripts/searchParamsCompiler/run.ts` as a convenient way to regenerate our schema files if we make changes the grammar/compiler.

Testing:
re-compiled the search params for fhir 4.0.1 and 3.0.1. There were no changes to the output, as expected.


src/implementationGuides/index.ts
```diff
@@ -0,0 +1,127 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+import { ImplementationGuides } from 'fhir-works-on-aws-interface';
```
nguyen102, line 6: Is this error a concern?

`Module '"../../node_modules/fhir-works-on-aws-interface/lib"' has no exported member 'ImplementationGuides'.`

I'm assuming it's because we need to update the `fhir-interface` that this package depends on?

src/implementationGuides/index.ts
```diff
@@ -0,0 +1,127 @@
...
+        typeof x.description === 'string' &&
+        Array.isArray(x.base) &&
+        x.base.every((y: any) => typeof y === 'string') &&
+        typeof x.type === 'string' &&
+        (x.expression === undefined || typeof x.expression === 'string') &&
+        (x.target === undefined || (Array.isArray(x.target) && x.target.every((y: any) => typeof y === 'string')))
+    );
+};
+
+const UNSUPPORTED_SEARCH_PARAMS = [
```
nguyen102, line 41: Should this be initialized as an array of string?

`const UNSUPPORTED_SEARCH_PARAMS: string[]`

nguyen102 (dismissed):
Looks good, one nitpick about type definition for  `UNSUPPORTED_SEARCH_PARAMS`. Also remember to update the version of the interface package. Looks like GH action Unit Tests and Linting threw an error for it.

src/implementationGuides/index.ts
carvantes: Doing so won't change the "initialization" of the variable. The difference would be declaring the type explicitly instead of letting TS infer it. 

In many cases declaring types explicitly makes code more readable, but declaring types explicitly all the time can be overly verbose and is sort of a waste of the type inference capabilities of TS.

IMO here it's obvious that it is an array of strings since it's a small, fixed, array declared on that same line.

src/implementationGuides/index.ts
carvantes: yeah, I just released the interface and bumped the version here and it's ok now.

src/implementationGuides/index.ts
```diff
@@ -0,0 +1,127 @@
...
+                target: searchParam.target,
+                compiled: searchParam.compiled.filter((x: any) => x.resourceType === base),
+            }));
+        });
+
+    return compiledSearchParams;
+};
+
+// eslint-disable-next-line import/prefer-default-export
+export const SearchImplementationGuides: ImplementationGuides = {
```
rsmayda, line 124: Can this compile method be used outside of the this search package (i.e. if I built a search off of an RDS instance)? should we have an IG/verification/compile package itself?

Read this comment https://github.com/awslabs/fhir-works-on-aws-interface/blob/mainline/src/implementationGuides.ts#L15; Seems to be each Search implementation will need to do its own compile

src/implementationGuides/index.ts
```diff
@@ -0,0 +1,127 @@
...
+    }
+    return true;
+};
+
+/**
+ * Compiles the contents of an Implementation Guide into an internal representation used to build Elasticsearch queries.
+ *
+ * @param searchParams - an array of FHIR SearchParameters. See: https://www.hl7.org/fhir/searchparameter.html
+ */
+const compile = async (searchParams: any[]): Promise<any> => {
```
rsmayda, line 79: Does this run at every Lambda invocation? any concerns with performance? What is the size of searchParams for USCore (for example)

src/implementationGuides/index.ts
```diff
@@ -0,0 +1,127 @@
...
+    const validFhirSearchParams: FhirSearchParam[] = [];
+    searchParams.forEach(s => {
+        if (isFhirSearchParam(s)) {
+            validFhirSearchParams.push(s);
```
rsmayda, line 83: Why are we passing in `any` type when we could just force the client to pass in the correct type `FhirSearchParam`?

src/implementationGuides/index.ts
```diff
@@ -0,0 +1,127 @@
...
+        } else {
+            throw new Error(`The following input is not a search parameter: ${JSON.stringify(s, null, 2)}`);
+        }
+    });
+
+    const compiledSearchParams = validFhirSearchParams
+        .filter(s => s.expression)
+        .filter(isParamSupported)
```
rsmayda: Should the `s.expression` call be a part of the `isParamSupported` method?

src/implementationGuides/index.ts
carvantes: No, It's not meant to run on every invocation.

The plan is to run it once at deploy time, The result is saved as JSON and included on the lambda deployment package.

src/implementationGuides/index.ts
carvantes: `FhirSearchParam` type is not exported, it is internal to this module. 
This method accepts `any[]` for convenience for the intended use case. Most of the time the client code won't know the exact type of the data, they'll just open the JSON files from the IGs and pass its contents to the compiler as is.

src/implementationGuides/index.ts
carvantes: good catch, fixed it. 
I actually had already included the condition on `isParamSupported` but forgot to remove this filter.

src/implementationGuides/index.ts
carvantes: Yes, it is expected that each Search Implementation does its own compile. 
Right now the output of this implementation is fairly generic and it COULD be re-used by other search implementations but we are not committing to keeping it free or ElasticSearch-specific stuff. This implementation is already doing custom stuff like excluding search params that cannot be supported by our ES implementation and we may later do things like generating ES mappings as part as the compile process once we stop relying on dynamic mappings. 

`ImplementationGuides` and `Search` interface are both implemented on this package but they don't truly depend on each other, so customers could import `fhir-works-on-aws-search-es` just for the IG compile part

```js
import { ImplementationGuides } from 'fhir-works-on-aws-search-es'
import { MySearch } from 'my-custom-es'
// do stuff
```

Although it's probably a good idea to later on extract the generic part (mostly the grammar parsing stuff) into a separate package so that customers can more easily reuse it when writing their `compile` implementation.

src/implementationGuides/index.ts
rsmayda: >  Most of the time the client code won't know the exact type of the data

Got it thanks!

rsmayda (approved):

PR #43, opened 2021-02-05
fix: properly handle special characters in queries

carvantes:
Description of changes:
Fixes part of https://github.com/awslabs/fhir-works-on-aws-deployment/issues/191
For `query_string` queries the following are reserved characters: `+ - = && || > < ! ( ) { } [ ] ^ " ~ * ? : \ /`. Not escaping them was causing issues.

Using `query_string` was not a great idea. `query_string` supports advanced syntax on `query` that we are not using at all and it's just causing issues. Right now we are treating all values simply as text so `multi_match` is a simpler query that gets the job done.

https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html

### DIFF
--- a/src/__snapshots__/elasticSearchService.test.ts.snap
+++ b/src/__snapshots__/elasticSearchService.test.ts.snap
@@ -1419,8 +1419,7 @@ Array [
             ],
             "must": Array [
               Object {
-                "query_string": Object {
-                  "default_operator": "AND",
+                "multi_match": Object {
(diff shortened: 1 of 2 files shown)


rsmayda (approved):
Looks good found this explainer on SO too:

https://stackoverflow.com/questions/37689935/whats-the-difference-between-query-string-and-multi-match

PR #44, opened 2021-02-05
feat: parse xpath expressions to support choice of data types

carvantes:
Description of changes:

Fixes part of https://github.com/awslabs/fhir-works-on-aws-deployment/issues/191

Supporting choice of data types using FHIRPath is hard because the type information is not contained in the search parameter expression. You'd need to somehow match it with the structure definition of the resources, and it gets trickier for profiles & implementation guides.

https://www.hl7.org/fhir/formats.html#choice

XPath is much better since the exact fields for the possible types are already contained on the xpath expression. However, we are not discarding FHIRPath entirely since there are conditions that can only be expressed in FHIRPath. Most notably resolving the type of a reference. i.e. `Invoice.subject.where(resolve() is Patient)`. So we are combining the results of parsing both expressions

i.e.
```
"expression": "ChargeItem.occurrence",
"xpath": "f:ChargeItem/f:occurrenceDateTime | f:ChargeItem/f:occurrencePeriod | f:ChargeItem/f:occurrenceTiming",
```

Note: the diff in the compiled search params is hard to see since they are minified to a single line. I added some intermediary commits where the JSON is formatted so that the diff is easier to see. You can check the `src/schema/compiledSearchParameters.4.0.1.json` file on
https://github.com/awslabs/fhir-works-on-aws-search-es/compare/d36a1aed48e516c3693d46f1b661259298396ed3...ef1aecc62250402651765866155bf977de035041 if you are curious.


src/implementationGuides/index.ts
```diff
@@ -5,7 +5,10 @@
 
 import { ImplementationGuides } from 'fhir-works-on-aws-interface';
 import * as nearley from 'nearley';
-import grammar from './reducedFHIRPath';
+import { Parser } from 'nearley';
```
rsmayda: do we need both of these imports?

src/implementationGuides/reducedXPath.ne
```diff
@@ -0,0 +1,35 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# Minimal subset of the XPath grammar required to parse the XPath SearchParameter expressions
+# See the docs for the syntax of this file at https://nearley.js.org/docs/grammar
+
+@preprocessor typescript
+
+Main -> unionExpression                                                              {% id %}
```
rsmayda, line 9: 1. Out of curiosity did you write these *.ne files or did you find time online? 
2. How do we test these? I guess right now is on the base R4/STU3 spec and then in the future IGs?

rsmayda:
> You can check the src/schema/compiledSearchParameters.4.0.1.json file on d36a1ae...ef1aecc if you are curious.

1. There is a huge change 26k lines on the STU3 file but only a three hundred of lines change on R4 -- why is that? any issues?
2. Looking at the compiled output I assume the `description` is there for the capabilityStatement?
3. I see that there is existing param `type`; could that be used to determine type information?
4. Could `compiled` param concat all paths into a single array:
```json
  "compiled": [
      {
        "resourceType": "DiagnosticReport",
        "path":[ "effective","effectiveDateTime", "effectivePeriod"]
      },
]
```

src/implementationGuides/reducedXPath.ne
carvantes: 1. I wrote them

2. There are unit tests for various cases, also it works for all the search params in R4/STU3 since we are compiling and committing the compiled files. I also tested with the 5 IGs in scope for CMS and it successfully parses all search params. Actually all of this started because  the search params for US Core were not working.

carvantes:
> > You can check the src/schema/compiledSearchParameters.4.0.1.json file on [d36a1ae...ef1aecc](https://github.com/awslabs/fhir-works-on-aws-search-es/compare/d36a1aed48e516c3693d46f1b661259298396ed3...ef1aecc62250402651765866155bf977de035041) if you are curious.
> 
> 1. There is a huge change 26k lines on the STU3 file but only a three hundred of lines change on R4 -- why is that? any issues?
> 2. Looking at the compiled output I assume the `description` is there for the capabilityStatement?
> 3. I see that there is existing param `type`; could that be used to determine type information?
> 4. Could `compiled` param concat all paths into a single array:
> 
> ```json
>   "compiled": [
>       {
>         "resourceType": "DiagnosticReport",
>         "path":[ "effective","effectiveDateTime", "effectivePeriod"]
>       },
> ]
> ```

1. No issues. It's just that I didn't pretty format the STU3 JSON file, so you see the change from going to 1 line to many. I only did the format thing for the R4 file.
2. yes. Description is just copied as is to be used as part of the capability statement.
3. Not really. `type` is the type of the search parameter, it is not necessarily the type of the field. There are fewer search param types than field types.(i.e. `Date` search type could map to fields of type `date`, `dateTime`, `instant`, `period`, etc. ). Also, `type` does not tell you if the field supports multiple types, so you don't know if you have to target it as `fieldName` or `fieldNameDateTime` for example.
4. That COULD be done in some cases where the compiled expressions do not have any conditions (or maybe they share an identical condition). But I see it as a low priority optimization.

src/implementationGuides/index.ts
carvantes: updated to a single line

rsmayda (approved):

PR #45, opened 2021-02-05
feat: accept compiled Implementation Guides as a parameter

carvantes:
Description of changes:

Accept compiled IGs as an optional parameter. Nothing changes if IGs are not being used


src/FHIRSearchParametersRegistry/index.ts
```diff
@@ -89,13 +89,16 @@ export class FHIRSearchParametersRegistry {
...
-    constructor(fhirVersion: FhirVersion) {
+    constructor(fhirVersion: FhirVersion, compiledImplementationGuides?: any[]) {
         let compiledSearchParams: SearchParam[];
         if (fhirVersion === '4.0.1') {
             compiledSearchParams = compiledSearchParamsV4 as SearchParam[];
         } else {
             compiledSearchParams = compiledSearchParamsV3 as SearchParam[];
         }
+        if (compiledImplementationGuides !== undefined) {
+            compiledSearchParams = [...compiledSearchParams, ...compiledImplementationGuides];
```
rsmayda, line 101: Concerns of duplicate params?

carvantes: Not really. If there are multiple params with the same `name`, the last one wins. We actually want the params from IGs to overwrite the base FHIR params in those cases. There are unit tests for that behavior.

I added a comment in the code to make this clearer

rsmayda (approved):

PR #54, opened 2021-04-07
feat: support prefixes and proper range queries for date parameters

carvantes:
Add support for date queries. https://www.hl7.org/fhir/search.html#date

key features are:
- prefixes
https://www.hl7.org/fhir/search.html#prefix
- treat partial dates as ranges
  > When the date parameter is not fully specified, matches against it are based on the behavior of intervals, where:
Dates with only the year specified are equivalent to an interval that starts at the first instant of January 1st to the last instant of December 31st, e.g. 2000 is equivalent to an interval of [2000-01-01T00:00, 2000-12-31T23:59].
...

### DIFF
--- a/src/QueryBuilder/typeQueries/dateQuery.test.ts
+++ b/src/QueryBuilder/typeQueries/dateQuery.test.ts
@@ -1,0 +1,225 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+import { InvalidSearchParameterError } from 'fhir-works-on-aws-interface';
+import each from 'jest-each';
+import { dateQuery, parseDateSearchParam } from './dateQuery';
+import { FHIRSearchParametersRegistry } from '../../FHIRSearchParametersRegistry';
+
+const fhirSearchParametersRegistry = new FHIRSearchParametersRegistry('4.0.1');
+const birthdateParam = fhirSearchParametersRegistry.getSearchParameter('Patient', 'birthdate')!.compiled[0];
+
+describe('parseDateSearchParam', () => {
+    describe('valid inputs', () => {
+        test('YYYY', () => {
+            expect(parseDateSearchParam('2020')).toMatchInlineSnapshot(`
+                            Object {
+                              "prefix": "eq",
+                              "range": Object {
+                                "end": 2020-12-31T23:59:59.000Z,
+                                "start": 2020-01-01T00:00:00.000Z,
+                              },
+                            }
+                    `);
+        });
+        test('YYYY-MM', () => {
+            expect(parseDateSearchParam('2020-02')).toMatchInlineSnapshot(`
+                            Object {
+                              "prefix": "eq",
+                              "range": Object {
+                                "end": 2020-02-29T23:59:59.000Z,
+                                "start": 2020-02-01T00:00:00.000Z,
+                              },
+                            }
+                    `);
+        });
+        test('YYYY-MM-DD', () => {
+            expect(parseDateSearchParam('2020-02-02')).toMatchInlineSnapshot(`
+                            Object {
+                              "prefix": "eq",
+                              "range": Object {
+                                "end": 2020-02-02T23:59:59.000Z,
+                                "start": 2020-02-02T00:00:00.000Z,
+                              },
+                            }
+                    `);
+        });
+        test('YYYY-MM-DDT:hh:mm', () => {
+            expect(parseDateSearchParam('2020-02-02T07:07')).toMatchInlineSnapshot(`
+                            Object {
+                              "prefix": "eq",
+                              "range": Object {
+                                "end": 2020-02-02T07:07:59.000Z,
+                                "start": 2020-02-02T07:07:00.000Z,
+                              },
+                            }
+                    `);
+        });
+        test('YYYY-MM-DDT:hh:mm:ss', () => {
+            expect(parseDateSearchParam('2020-02-02T07:07:07')).toMatchInlineSnapshot(`
+                            Object {
+                              "prefix": "eq",
+                              "range": Object {
+                                "end": 2020-02-02T07:07:07.000Z,
+                                "start": 2020-02-02T07:07:07.000Z,
+                              },
+                            }
+                    `);
+        });
+        test('YYYY-MM-DDT:hh:mm:ssZ', () => {
+            expect(parseDateSearchParam('2020-02-02T07:07:07Z')).toMatchInlineSnapshot(`
+                Object {
+                  "prefix": "eq",
+                  "range": Object {
+                    "end": 2020-02-02T07:07:07.000Z,
+                    "start": 2020-02-02T07:07:07.000Z,
+                  },
+                }
+            `);
+        });
+        test('YYYY-MM-DDT:hh:mm:ss+hh:mm', () => {
+            expect(parseDateSearchParam('2020-02-02T07:07:07+07:00')).toMatchInlineSnapshot(`
+                Object {
+                  "prefix": "eq",
+                  "range": Object {
+                    "end": 2020-02-02T00:07:07.000Z,
+                    "start": 2020-02-02T00:07:07.000Z,
+                  },
+                }
+            `);
+        });
+    });
+
+    describe('invalid inputs', () => {
+        each([
+            ['badpre2020-02-02'],
+            ['This is not a date at all'],
+            ['2020-99'],
+            ['2020-99-99'],
+            ['2020/02/02'],
+            ['2020-02-02T07'],
+            ['2020-02-02T07:07:07someSuffix'],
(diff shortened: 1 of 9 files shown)


src/QueryBuilder/typeQueries/dateQuery.ts
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// The date parameter format is yyyy-mm-ddThh:mm:ss[Z|(+|-)hh:mm] (the standard XML format).
+// https://www.hl7.org/fhir/search.html#date
+const DATE_SEARCH_PARAM_REGEX = /^(?<prefix>eq|ne|lt|gt|ge|le|sa|eb|ap)?(?<inputDate>(?<year>\d{4})(?:-(?<month>\d{2})(?:-(?<day>\d{2})(?:T(?<hours>\d{2}):(?<minutes>\d{2})(?::(?<seconds>\d{2})(?<timeZone>Z|[+-](?:\d{2}:\d{2}))?)?)?)?)?)$/;
+
+export const parseDateSearchParam = (param: string): DateSearchParameter => {
+    const match = param.match(DATE_SEARCH_PARAM_REGEX);
+    if (match === null) {
+        throw new InvalidSearchParameterError(`Invalid date search parameter: ${param}`);
```
rsmayda, line 29: Just to clarify from a user experience perspective if they submit an invalid date the search will return a 4xx request error?

src/QueryBuilder/typeQueries/dateQuery.ts
```diff
@@ -0,0 +1,149 @@
...
+    }
+    const { inputDate, month, day, minutes, seconds } = match.groups!;
+
+    // If no prefix is present, the prefix eq is assumed.
+    // https://www.hl7.org/fhir/search.html#prefix
+    const prefix = match.groups!.prefix ?? 'eq';
+
+    const parsedDate = parseISO(inputDate);
+    if (!isValid(parsedDate)) {
```
rsmayda, line 38: what's an example of an invalid date that passes regex? I think like 2020-21-90?

src/QueryBuilder/typeQueries/dateQuery.ts
```diff
@@ -0,0 +1,149 @@
...
+    if (match === null) {
+        throw new InvalidSearchParameterError(`Invalid date search parameter: ${param}`);
+    }
+    const { inputDate, month, day, minutes, seconds } = match.groups!;
+
+    // If no prefix is present, the prefix eq is assumed.
+    // https://www.hl7.org/fhir/search.html#prefix
+    const prefix = match.groups!.prefix ?? 'eq';
+
+    const parsedDate = parseISO(inputDate);
```
rsmayda, line 37: This will convert "2021-03" to "2021-03-01<timestamp here>" just want to make sure the start and end dates returned are full dates

src/QueryBuilder/typeQueries/dateQuery.ts
```diff
@@ -0,0 +1,149 @@
...
+            start: parsedDate,
+            end: endDate,
+        },
+    };
+};
+
+// eslint-disable-next-line import/prefer-default-export
+export const dateQuery = (compiledSearchParam: CompiledSearchParam, value: string): any => {
+    const dateSearchParameter = parseDateSearchParam(value);
+    const { start, end } = dateSearchParameter.range;
```
rsmayda: nit: 
```suggestion
    const { prefix, range}  = parseDateSearchParam(value);
    const { start, end } = range;
```

src/QueryBuilder/typeQueries/dateQuery.ts
```diff
@@ -0,0 +1,149 @@
...
+
+    // See https://www.hl7.org/fhir/search.html#prefix
+    if (dateSearchParameter.prefix !== 'ne') {
```
rsmayda: So if the prefix is 'ne' we return nothing?

src/QueryBuilder/typeQueries/dateQuery.ts
```diff
@@ -0,0 +1,149 @@
...
+            range: {
+                [compiledSearchParam.path]: elasticSearchRange,
+            },
+        };
+    }
+
+    // ne prefix is the only case that requires a bool query;
+    const neQuery = {
+        bool: {
+            should: [
```
rsmayda, line 129: So is the should an OR query? date must be greater than end OR date must be less than start

helpful link I found: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-bool-query.html

src/QueryBuilder/typeQueries/dateQuery.ts
carvantes: yes, `InvalidSearchParameterError ` is mapped to a 400 error

src/QueryBuilder/typeQueries/dateQuery.test.ts
```diff
@@ -0,0 +1,225 @@
...
+                            Object {
+                              "prefix": "eq",
+                              "range": Object {
+                                "end": 2020-02-02T07:07:07.000Z,
+                                "start": 2020-02-02T07:07:07.000Z,
+                              },
+                            }
+                    `);
+        });
+        test('YYYY-MM-DDT:hh:mm:ssZ', () => {
```
rsmayda, line 70: Not putting the +/d/dZ does this default to UTC?

src/QueryBuilder/typeQueries/dateQuery.test.ts
```diff
@@ -0,0 +1,225 @@
...
+            ['2020-02-02T07:07:07someSuffix'],
+            ['2020-02-02someSuffix'],
+        ]).test('%s', param => {
+            expect(() => parseDateSearchParam(param)).toThrow(InvalidSearchParameterError);
+        });
+    });
+});
+
+describe('buildDateQuery', () => {
+    test('np prefix', () => {
```
rsmayda:
```suggestion
    test('no prefix', () => {
```

src/elasticSearchService.test.ts
```diff
@@ -42,6 +42,7 @@ describe('typeSearch', () => {
             [{}],
             [{ _count: 10, _getpagesoffset: 2 }],
             [{ gender: 'female', name: 'Emily' }],
+            [{ gender: 'female', birthdate: 'gt1990' }],
```
rsmayda, line 45: How are dates currently being saved in ES/DDB? Or is this a separate issue?

src/QueryBuilder/typeQueries/dateQuery.ts
carvantes: Your example is correct. mostly dates with values out of bounds. there are some other examples in unit tests like `2020-99-99`

src/QueryBuilder/typeQueries/dateQuery.ts
carvantes: `2021-03` is parsed into `2021-03-01T00:00:00.000Z`. ISO dates fill missing values(months, days, hours, etc,) with the lowest value possible. That allows us to use it as the start date. 

There are several unit tests that show the behavior for partial dates.

src/QueryBuilder/typeQueries/dateQuery.ts
carvantes: `ne` prefix is handled at L127. It's just convenient to handle it separately since it's the only prefix that results in a different query structure

src/QueryBuilder/typeQueries/dateQuery.ts
carvantes: yes, `should` behaves like an `OR` when it's the only clause in a `bool` query

src/QueryBuilder/typeQueries/dateQuery.test.ts
carvantes: It defaults to the server time. In our case it ends up as UTC. Lambda uses UTC regardless of the region it's deployed

src/QueryBuilder/typeQueries/dateQuery.ts
nguyen102: Based on this link 
`should` is equivalent to `OR`
`must` is equivalent to `AND`
https://stackoverflow.com/a/40755927/14310364

src/elasticSearchService.test.ts
carvantes: in DB they are merely strings. In ES they are indexed as Date fields. 

https://www.elastic.co/guide/en/elasticsearch/reference/current/date.html

src/QueryBuilder/typeQueries/dateQuery.test.ts
carvantes: updated

src/QueryBuilder/typeQueries/dateQuery.ts
carvantes: updated

rsmayda (approved):

PR #56, opened 2021-04-09
feat: support token parameters

carvantes:
Add support for token queries: https://www.hl7.org/fhir/search.html#token

the key feature is supporting `|` to indicate system and code:

> - **[parameter]=[code]**: the value of [code] matches a Coding.code or Identifier.value irrespective of the value of the system property
> - **[parameter]=[system]|[code]**: the value of [code] matches a Coding.code or Identifier.value, and the value of [system] matches the system property of the Identifier or Coding
> - **[parameter]=|[code]**: the value of [code] matches a Coding.code or Identifier.value, and the Coding/Identifier has no system property
> - **[parameter]=[system]|**: any element where the value of [system] matches the system property of the Identifier or Coding

### DIFF
--- a/src/QueryBuilder/typeQueries/tokenQuery.test.ts
+++ b/src/QueryBuilder/typeQueries/tokenQuery.test.ts
@@ -1,0 +1,186 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+import { InvalidSearchParameterError } from 'fhir-works-on-aws-interface';
+import { parseTokenSearchParam, tokenQuery } from './tokenQuery';
+import { FHIRSearchParametersRegistry } from '../../FHIRSearchParametersRegistry';
+
+const fhirSearchParametersRegistry = new FHIRSearchParametersRegistry('4.0.1');
+const identifierParam = fhirSearchParametersRegistry.getSearchParameter('Patient', 'identifier')!.compiled[0];
+
+describe('parseTokenSearchParam', () => {
+    describe('valid inputs', () => {
+        test('code', () => {
+            expect(parseTokenSearchParam('code')).toMatchInlineSnapshot(`
+                Object {
+                  "code": "code",
+                  "explicitNoSystemProperty": false,
+                  "system": undefined,
+                }
+            `);
+        });
+        test('system|code', () => {
+            expect(parseTokenSearchParam('system|code')).toMatchInlineSnapshot(`
+                Object {
+                  "code": "code",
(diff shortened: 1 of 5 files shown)


src/QueryBuilder/typeQueries/tokenQuery.ts
```diff
@@ -0,0 +1,100 @@
...
+        [code] = parts;
+    } else {
+        [system, code] = parts;
+        if (system === '') {
+            system = undefined;
+            explicitNoSystemProperty = true;
+        }
+    }
+    code = code === '' ? undefined : code;
+    return { system, code, explicitNoSystemProperty };
```
rsmayda, line 39: This can return {undefined, undefined, false}

src/QueryBuilder/typeQueries/tokenQuery.ts
```diff
@@ -0,0 +1,100 @@
...
+        });
+    }
+
+    if (queries.length === 1) {
+        return queries[0];
+    }
+
+    return {
+        bool: {
+            must: queries,
```
rsmayda, line 102: I think the use-case with param either `''` or `'|'` can happen and will result in `queries = []` is that okay?

src/QueryBuilder/typeQueries/tokenQuery.test.ts
```diff
@@ -0,0 +1,186 @@
...
+                Object {
+                  "code": "2345",
+                  "explicitNoSystemProperty": false,
+                  "system": "http://acme.org/patient",
+                }
+            `);
+        });
+    });
+
+    describe('invalid inputs', () => {
```
rsmayda, line 71: add '' and '|' test cases (unless they are valid?)

rsmayda (changes requested):

src/QueryBuilder/typeQueries/tokenQuery.test.ts
carvantes: added tests

src/QueryBuilder/typeQueries/tokenQuery.ts
carvantes: I made a few changes. 
- `''` is valid, it ends up as searching for an empty code in any system(not a very useful query, but it's valid anyways)
- `|` is now explicitly an invalid param. I don't think that it can be unambiguously interpreted per the FHIR spec. HAPI allows such query but it seems to never match anything. I think it makes sense to explicitly reject it

src/QueryBuilder/typeQueries/tokenQuery.ts
carvantes: updated

rsmayda (approved):

PR #58, opened 2021-04-16
feat: support number and quantity type parameters

carvantes:
Description of changes:

`number` and `quantity` search param types are similar. `quantity` is essentially a `number` with units(expressed via system + code)

https://www.hl7.org/fhir/search.html#number
https://www.hl7.org/fhir/search.html#quantity

key features:
- Implicit range according to number of significant digits.
  >Searching on decimals involves an implicit range. The number of significant digits of the implicit range is the number of digits specified in the search parameter value, excluding leading zeros. So 100 and 1.00e2 both have the same number of significant digits - three
  >
  > [parameter]=100.00 | Values that equal 100, to 5 significant figures precision, so this is actually searching for values in the range [99.995 ... 100.005)

- Quantity type allows to specify `system` and `code`
- Prefixes. Behavior is mostly the same as the prefixes for Dates. Some refactoring was done to reuse prefixes code across several query types.


src/QueryBuilder/typeQueries/common/prefixRange.ts
```diff
@@ -0,0 +1,124 @@
...
+
+        return {
+            range: {
+                [path]: elasticSearchRange,
+            },
+        };
+    }
+
+    // ne prefix is the only case that requires a bool query;
+    const neQuery = {
```
nguyen102: Can we put this at the beginning of the method? This will allow us to "return" earlier. Something along the lines of 
```
if (prefix === 'ne') {
    return neQuery;
}
// Handle other cases

```

src/QueryBuilder/typeQueries/quantityQuery.test.ts
```diff
@@ -0,0 +1,177 @@
...
+        });
+        test('5.40e-3|http://unitsofmeasure.org|g', () => {
+            expect(quantityQuery(quantityParam, '5.40e-3|http://unitsofmeasure.org|g')).toMatchInlineSnapshot(`
+                Object {
+                  "bool": Object {
+                    "must": Array [
+                      Object {
+                        "range": Object {
+                          "valueQuantity.value": Object {
+                            "gte": 0.0053950000000000005,
```
nguyen102, line 60: Are all of the `0`s before the 5 intended? Why not just `0.005395`?

nguyen102 (commented):
Overall looks good. One minor change suggestion, and one question.

src/QueryBuilder/typeQueries/common/prefixRange.ts
carvantes: I like that idea. I made the change

src/QueryBuilder/typeQueries/quantityQuery.test.ts
carvantes: In practical terms, the number is `0.005395`. The `...00005` is rounding error due to floating-point arithmetic. This is fairly standard  and is one of the reasons why FHIR proposes comparing against a small range around the number instead of doing equality comparisons for number searches.

https://floating-point-gui.de/

nguyen102 (approved):

PR #60, opened 2021-04-21
feat: allow sorting by date type parameters

carvantes:
see: https://www.hl7.org/fhir/search.html#sort

example: `GET [base]/Observation?_sort=status,-date,category`

only date is supported for now since that's the only one required to unblock `$docref`(plus, it may be the most useful for customers), I'll add a backlog item to support the other types.

**Edit:** updated to also work with fields of type Period (they have a `start` and an `end`) turns out that $docref needs that

### DIFF
--- a/src/QueryBuilder/index.ts
+++ b/src/QueryBuilder/index.ts
@@ -111,3 +111,5 @@ export const buildQueryForAllSearchParameters = (
         },
     };
 };
+
+export { buildSortClause } from './sort';
--- a/src/QueryBuilder/sort.ts
+++ b/src/QueryBuilder/sort.ts
@@ -23,12 +23,21 @@ export const parseSortParameter = (param: string): SortParameter[] => {
     });
 };
 
+const elasticsearchSort = (field: string, order: 'asc' | 'desc') => ({
+    [field]: {
+        order,
+        // unmapped_type makes queries more fault tolerant. Since we are using dynamic mapping there's no guarantee
+        // that the mapping exists at query time. This ignores the unmapped field instead of failing
+        unmapped_type: 'long',
+    },
+});
+
 // eslint-disable-next-line import/prefer-default-export
 export const buildSortClause = (
     fhirSearchParametersRegistry: FHIRSearchParametersRegistry,
     resourceType: string,
     sortQueryParam: string | string[],
-): any => {
+): any[] => {
     if (Array.isArray(sortQueryParam)) {
         throw new InvalidSearchParameterError('_sort parameter cannot be used multiple times on a search query');
     }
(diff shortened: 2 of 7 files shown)


src/QueryBuilder/index.ts
```diff
@@ -111,3 +111,5 @@ export const buildQueryForAllSearchParameters = (
         },
     };
 };
+
+export { buildSortClause } from './sort';
```
rsmayda, line 115: What is this line doing? just exporting a method from ./sort file? Why would we want to reexport this method as it's own?

src/QueryBuilder/sort.ts
```diff
@@ -0,0 +1,72 @@
...
+    return sortParams.flatMap(sortParam => {
+        const searchParameter = fhirSearchParametersRegistry.getSearchParameter(resourceType, sortParam.searchParam);
+        if (searchParameter === undefined) {
+            throw new InvalidSearchParameterError(
+                `Unknown _sort parameter value: ${sortParam.searchParam}. Sort parameters values must use a valid Search Parameter`,
+            );
+        }
+        if (searchParameter.type !== 'date') {
+            throw new InvalidSearchParameterError(
+                `Invalid _sort parameter: ${sortParam.searchParam}. Only date type parameters can be used for sorting`,
```
rsmayda: Should we re-word this to be "Only date type parameters can **currently** be used for sorting"? This is a system gap not necessarily a FHIR spec one

src/QueryBuilder/sort.ts
```diff
@@ -0,0 +1,72 @@
...
+        return searchParameter.compiled.flatMap(compiledParam => {
+            return [
+                elasticsearchSort(compiledParam.path, sortParam.order),
+
+                // Date search params may target fields of type Period, so we add a sort clause for them.
+                // The FHIR spec does not fully specify how to sort by Period, but it makes sense that the most recent
+                // record is the one with the most recent "end" date and vice versa.
+                elasticsearchSort(
+                    sortParam.order === 'desc' ? `${compiledParam.path}.end` : `${compiledParam.path}.start`,
+                    sortParam.order,
```
rsmayda, line 67: Should the Period sorting always be included?

rsmayda (commented):
Very minor questions/comments

src/QueryBuilder/sort.ts
carvantes: same reason as https://github.com/awslabs/fhir-works-on-aws-search-es/pull/61#discussion_r618713667

src/QueryBuilder/sort.ts
carvantes: I like that. Updated

src/QueryBuilder/index.ts
carvantes: This is just to keep a clean interface for `QueryBuilder`. outsiders should only import stuff from `./QueryBuilder` and not peek at its internal functions/classes. i.e importing `./QueryBuilder/typeQueries/numberQuery` is discouraged

rsmayda (approved):

PR #61, opened 2021-04-22
feat: support Period type fields for date params

carvantes:
Enhance the date queries to work with fields of type Period(they have `start` and `end` datetime values). This means that queries are matching the range of the period against the implicit range of the date parameter.

See the interpretation of prefixes when applied to ranges: https://www.hl7.org/fhir/search.html#prefix. It can be counterintuitive at first

**Note:** we rely on dynamic mapping and we don't know always know if a given field is a date or a period, so queries include clauses for both. clauses for invalid fields are simply ignored. The query results are correct, but queries can be simplified once we have static mapping and know the exact type of all fields.

### DIFF
--- a/src/QueryBuilder/typeQueries/common/prefixRange.ts
+++ b/src/QueryBuilder/typeQueries/common/prefixRange.ts
@@ -3,8 +3,6 @@
  *  SPDX-License-Identifier: Apache-2.0
  */
 
-import { InvalidSearchParameterError } from 'fhir-works-on-aws-interface';
-
 interface Range {
     start: Date | number;
     end: Date | number;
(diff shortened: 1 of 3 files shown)


src/QueryBuilder/typeQueries/common/prefixRange.ts
```diff
@@ -20,6 +18,91 @@ interface DateRange {
...
+        case 'eq':
+            return eqQuery;
+        case 'ne':
+            return {
+                bool: {
+                    must_not: eqQuery,
+                },
+            };
+        case 'lt':
+        case 'le':
```
rsmayda, line 73: Should these return the same? lt == less than le == less than or equal to (same for ge/gt)

Wording from FHIR is:

> The range before 2015-08-12T05:23:45 includes any time up to 2015-08-12T05:23:45.000000000000000
--

src/QueryBuilder/typeQueries/common/prefixRange.ts
```diff
@@ -20,6 +18,91 @@ interface DateRange {
...
+        case 'eb':
+            return ebQuery;
+        case 'sa':
+            return saQuery;
+        case 'ap':
+            return {
+                bool: {
+                    must_not: [ebQuery, saQuery],
+                },
+            };
```
rsmayda, line 99: What does this mean from a ES search? the end value must NOT end before start AND the start value must NOT start after end?

src/QueryBuilder/typeQueries/common/prefixRange.ts
```diff
@@ -116,5 +204,12 @@ export const prefixRangeNumber = (prefix: string, number: number, implicitRange:
 };
 
 export const prefixRangeDate = (prefix: string, range: DateRange, path: string): any => {
-    return prefixRange(prefix, range, path);
+    return {
+        bool: {
+            should: [
+                prefixRange(prefix, range, path), // date, dateTime, instant
+                prefixRangePeriod(prefix, range, path), // Period
```
rsmayda, line 211: Should we always include this?

src/QueryBuilder/typeQueries/common/prefixRange.ts
carvantes: `ap` means "the range of the search value overlaps with the range of the target value". 
`eb` and `sa` are the queries for disjointed ranges. It's shorter to express `ap` by negating `eb` and `sa` (If it is not disjoint, then it overlaps)

src/QueryBuilder/typeQueries/common/prefixRange.ts
carvantes: for now, yes. 
> Note: we rely on dynamic mapping and we don't know always know if a given field is a date or a period, so queries include clauses for both. clauses for invalid fields are simply ignored. The query results are correct, but queries can be simplified once we have static mapping and know the exact type of all fields.

src/QueryBuilder/typeQueries/common/prefixRange.ts
carvantes: afaik it is the same.
The wording you mention is for the range before a value. It's not exactly the same as the range before a range.

the thing is that `eq` for ranges means that "the range of the search value fully contains the range of the target value" this is actually a very narrow condition. And if a range contains another range, then it is implicitly already less than(and greater than) it. So the eq part of lte is redundant.

rsmayda: ooo interesting thank you

src/QueryBuilder/typeQueries/common/prefixRange.ts
rsmayda: d'oh thanks

rsmayda (approved):

rsmayda (approved):

PR #62, opened 2021-04-23
feat: allow repeated search parameters

carvantes:
https://www.hl7.org/fhir/search.html#combining

If a parameter repeats, such as `/Patient?language=FR&language=NL`, then this matches a patient who speaks both languages. This is known as an AND search parameter, since the server is expected to respond only with results which match both values.

Change is actually fairly small since multiple different search parameters are combined with AND anyways

### DIFF
--- a/src/__snapshots__/elasticSearchService.test.ts.snap
+++ b/src/__snapshots__/elasticSearchService.test.ts.snap
@@ -1403,7 +1403,7 @@ Array [
 ]
 `;
 
-exports[`typeSearch query snapshots for simple queryParams; with ACTIVE filter queryParams={"_count":10,"_getpagesoffset":2,"_id":"11111111-1111-1111-1111-111111111111","gender":"female","name":"Emily","_format":"json"} 1`] = `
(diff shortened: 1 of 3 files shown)


src/__snapshots__/elasticSearchService.test.ts.snap
```diff
@@ -1751,6 +1751,66 @@ Array [
 ]
 `;
 
+exports[`typeSearch query snapshots for simple queryParams; with ACTIVE filter queryParams={"gender":"female","name":["Emily","Smith"]} 1`] = `
```
nguyen102, line 1754: Is there suppose to be a `1` on this line? What does that represent?

carvantes: you mean the `1` at the end?

That's added by jest for all snapshot names. All the other snapshots have it as well.

https://jestjs.io/docs/snapshot-testing#snapshot-testing-with-jest

nguyen102 (approved):

PR #64, opened 2021-04-23
fix: fix ne prefix was matching extra records

carvantes:
The recent addition of support for Period type fields broke search for regular date fields when using the `ne` prefix.
queries with `ne` where returning additional results and integ tests caught the issue.

The issue was a bad combination of the implicit failure of non-existent fields in range queries + must_not clauses

All integ test pass after this fix. 


src/QueryBuilder/typeQueries/common/prefixRange.ts
```diff
@@ -59,48 +59,77 @@ const prefixRangePeriod = (prefix: string, range: Range, periodTypeField: string
...
+                // especially when must_not clauses are involved
+                {
+                    exists: {
+                        field: startField,
+                    },
+                },
+                {
+                    exists: {
+                        field: endField,
+                    },
```
rsmayda, line 127: Is it valid to have a start period but no end (or vice versa)?

rsmayda: https://www.hl7.org/fhir/datatypes.html#period

FYI it looks like it is a usecase -- how much does it complicate things?

carvantes: yeah, I saw that start and end are optional. 

The meaning of missing bounds in FHIR is rather vague and that's why I didn't immediately addressed that case:
_If the start element is missing, the start of the period is not known(so what?). If the end element is missing, it means that the period is ongoing(does "ongoing" extends until now? or infinitely into the future?)_

My interpretation would be that no start extends infinitely into the past and no end extends infinitely into the future. I'm not sure about how to validate it, but from the source code it looks like the [IBM implementation](https://github.com/IBM/FHIR/blob/a2a98d49c431dc8c94bc5488fc1a446c6bd5bdcb/fhir-persistence/src/test/java/com/ibm/fhir/persistence/search/test/AbstractSearchDateTest.java#L1766) agrees with that interpretation, so we'd likely go with that.

In any case, those cases can be part of a separate PR. They won't change the current query but rather expand it via composition.

Currently we have:
```
(start exists && end exists && prefixRangePeriodQuery)
```

The goal would be: 
```
(start exists && end exists && prefixRangePeriodQuery)
|| (start not exists && end exists && prefixRangePeriodNoStartQuery)
|| (start exists && end not exists && prefixRangePeriodNoEndQuery)
```

rsmayda (approved):

PR #65, opened 2021-04-28
fix: token search params were matching additional documents

### LINKED ISSUE #301: Not able to search resource identifier exactly.
Hi FHIR Works Team,
I tried to:
`GET /Patient?identifier=testuser01@gmail.com`
and there is only one patient with `identifier = testuser01@gmail.com`.
I expect it to return one result, but it returns many more.

Is this a bug?
Deployed fhir-works-on-aws-deployment version: v2.6.0

carvantes:
Issue #, if available:
Fix https://github.com/awslabs/fhir-works-on-aws-deployment/issues/301

Description of changes:

Turns out that the identifier param was resulting in text search queries instead of exact text matching.
e.g. testuser01@gmail.com was analyzed into words before searching: testuser0,gmail.com. This ended up matching unwanted records with a gmail.com address.

Always using the `.value.keyword` field fixes the issue


rsmayda (approved):
Could we add an integ test for this use case?

PR #71, opened 2021-05-14
refactor: add custom logger

Bingjiling:
Issue #, if available:

Description of changes:

### DIFF
--- a/src/loggerBuilder.ts
+++ b/src/loggerBuilder.ts
@@ -1,0 +1,9 @@
+import { makeLogger } from 'fhir-works-on-aws-interface';
+
+const componentLogger = makeLogger({
+    component: 'routing',
+});
+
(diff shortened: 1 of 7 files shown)


src/loggerBuilder.ts
```diff
@@ -0,0 +1,9 @@
+import { makeLogger } from 'fhir-works-on-aws-interface';
+
+const componentLogger = makeLogger({
+    component: 'routing',
```
carvantes:
```suggestion
    component: 'search',
```

carvantes (changes requested):

src/loggerBuilder.ts
Bingjiling: Good catch!

carvantes (approved):

PR #76, opened 2021-05-28
fix: Show true size for ES results

rsmayda:
Description of changes:

* If ES has more than 10k results in the search the count will only say 10k. This change returns the true value. 
* If someone tries to access > the window limit we will throw a 400 invalid parameters instead of 500

### DIFF
--- a/src/elasticSearchService.ts
+++ b/src/elasticSearchService.ts
@@ -17,6 +17,7 @@ import {
     SearchEntry,
     SearchFilter,
     FhirVersion,
+    InvalidSearchParameterError,
 } from 'fhir-works-on-aws-interface';
 import { Client } from '@elastic/elasticsearch';
 import { ElasticSearch } from './elasticSearch';
@@ -102,6 +103,7 @@ export class ElasticSearchService implements Search {
                 index: resourceType.toLowerCase(),
                 from,
                 size,
+                track_total_hits: true,
                 body: {
                     query,
                 },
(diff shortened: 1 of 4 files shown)


src/elasticSearchService.ts
```diff
@@ -175,6 +177,14 @@ export class ElasticSearchService implements Search {
                     hits: [],
                 };
             }
+            if (
+                error instanceof ResponseError &&
+                error.message === 'search_phase_execution_exception' &&
+                error.body.error.caused_by.type === 'illegal_argument_exception'
+            ) {
```
rsmayda: This seems frail (and may be ES version dependent).  I almost think it is better for us to return a 500 internal server error -- thoughts?

carvantes: I'm not sure about it being frail, but I don't like that the >10k results error can overlap with other `illegal_argument_exception` errors. 

I think it would be better to do explicit validation on our code on parameters such as `_count`, and `_getpagesoffset` to make sure that they are within the bounds that we can process.

That way we can confidently return a 400 error with a more specific message like _"Unable to return more than 10k results. Try using more specific search parameters to narrow down the search results"_

rsmayda: Yeah I thought about do that but implementations can change their window size to be greater than or less than 10k -- perhaps we can pass a param to ES constructor?

"By default the offset + limit is limited to 10,000. This can be modified at a cluster level, but I would seriously advise against doing do."
https://discuss.elastic.co/t/result-window-is-too-large-from-size-must-be-less-than-or-equal-to-10000-but-was-10050/124392

carvantes: I wouldn't make it a constructor param. I feel like it won't ever be used.
I think it's ok to just have a const value with the limit, we can always add it as constructor param later if there's customer feedback.

afaik it is rare to change the `max_result_window` value in ES. Increasing it can cause performance issues and clients who need more than 10k results should be using the scroll API instead. Lowering it is rarely done since 10k is a sensible default (ES puts a lot of thought on the default settings)

carvantes (approved):

PR #79, opened 2021-06-11
feat!: Use alias in place of index for search

Bingjiling:
Issue #, if available:

Description of changes:

* Update search to use alias in place of index 
* Update snapshots 
* BREAKING CHANGE: alias need to be added to existing index 
* Deployed to AWS account and verified all integ tests passes 

### DIFF
--- a/src/__snapshots__/elasticSearchService.test.ts.snap
+++ b/src/__snapshots__/elasticSearchService.test.ts.snap
@@ -1,6 +1,84 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`typeSearch _include queryParams={"_include":"*"}: msearch queries 1`] = `Array []`;
+exports[`typeSearch _include queryParams={"_include":"*"}: msearch queries 1`] = `
+Array [
+  Array [
+    Object {
+      "body": Array [
+        Object {
+          "index": "patient-alias",
+        },
+        Object {
+          "query": Object {
+            "bool": Object {
+              "filter": Array [
+                Object {
+                  "terms": Object {
+                    "id": Array [
+                      "patient-id-333",
+                    ],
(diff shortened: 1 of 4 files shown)


src/searchInclusions.ts
```diff
@@ -174,7 +174,7 @@ export const buildIncludeQuery = (
     resourceIds: string[],
     filterRulesForActiveResources: any[],
 ) => ({
-    index: resourceType.toLowerCase(),
```
carvantes, line 177: A similar change is also needed on: https://github.com/awslabs/fhir-works-on-aws-search-es/blob/e5bf19d6a4f7a519b4ba7b67d27f95c0527baff0/src/searchInclusions.ts#L201

src/__snapshots__/elasticSearchService.test.ts.snap
```diff
@@ -1,84 +1,6 @@
...
-              ],
-            },
-          },
-        },
-      ],
-    },
-  ],
-]
-`;
+exports[`typeSearch _include queryParams={"_include":"*"}: msearch queries 1`] = `Array []`;
```
carvantes: Looks like the changes broke this test. The expected change was only to append `-alias`, to the `index`. The snapshot shouldn't be an empty array. 

I think this breaks since some filter logic assumed that index was a valid resourceType:

https://github.com/awslabs/fhir-works-on-aws-search-es/blob/e5bf19d6a4f7a519b4ba7b67d27f95c0527baff0/src/elasticSearchService.ts#L286-L289

keep an eye for other places where index was referenced and treated as a resourceType(I don't think that there are many more).

carvantes (changes requested):

carvantes (approved):

PR #84, opened 2021-06-28
feat: implement multi-tenancy

carvantes:
When multi-tenancy is enabled, a filter by `_tenantId` clause is added to ALL queries. Also queries use the tenant-specific alias.

Related PR: https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/pull/91

### DIFF
--- a/src/__snapshots__/elasticSearchService.test.ts.snap
+++ b/src/__snapshots__/elasticSearchService.test.ts.snap
@@ -1299,6 +1299,316 @@ Array [
 ]
 `;
 
+exports[`typeSearch multi-tenancy enabled: _include:iterate: msearch queries 1`] = `
+Array [
+  Array [
+    Object {
+      "body": Array [
+        Object {
+          "index": "patient-alias-tenant-tenant1",
+        },
+        Object {
+          "query": Object {
+            "bool": Object {
+              "filter": Array [
+                Object {
+                  "terms": Object {
+                    "id": Array [
+                      "patient-id-333",
+                    ],
+                  },
+                },
+                Object {
+                  "match": Object {
+                    "someFieldThatTellsIfTheResourceIsActive": "AVAILABLE",
+                  },
+                },
+                Object {
+                  "match": Object {
+                    "_tenantId.keyword": "tenant1",
+                  },
+                },
+              ],
+            },
+          },
+        },
+      ],
+    },
+  ],
+  Array [
+    Object {
+      "body": Array [
+        Object {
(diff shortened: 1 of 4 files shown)


src/elasticSearchService.ts
```diff
@@ -68,23 +84,53 @@ export class ElasticSearchService implements Search {
...
+        const { searchFilters, tenantId } = request;
+        const filters: any[] = ElasticSearchService.buildElasticSearchFilter([
+            ...this.searchFiltersForAllQueries,
+            ...(searchFilters ?? []),
+        ]);
+
+        if (this.enableMultiTenancy) {
+            filters.push({
+                match: {
+                    '_tenantId.keyword': tenantId,
```
rsmayda, line 120: If "_tenantId" is a keyword type can we still use ".keyword"?

src/elasticSearchService.ts
```diff
@@ -172,20 +218,30 @@ export class ElasticSearchService implements Search {
...
             return {
                 total: apiResponse.body.hits.total.value,
                 hits: apiResponse.body.hits.hits,
             };
         } catch (error) {
             // Indexes are created the first time a resource of a given type is written to DDB.
             if (error instanceof ResponseError && error.message === 'index_not_found_exception') {
-                logger.info(`Search index for ${searchQuery.index} does not exist. Returning an empty search result`);
+                logger.info(
+                    `Search index for ${searchQuery.queryRequest.index} does not exist. Returning an empty search result`,
```
rsmayda, line 243: Do we still get this error when searching off of aliases?

rsmayda (approved):

src/__snapshots__/elasticSearchService.test.ts.snap
```diff
@@ -1299,6 +1299,316 @@ Array [
...
+                },
+                Object {
+                  "match": Object {
+                    "someFieldThatTellsIfTheResourceIsActive": "AVAILABLE",
+                  },
+                },
+                Object {
+                  "match": Object {
+                    "_tenantId.keyword": "tenant1",
+                  },
```
rsmayda, line 1329: This isn't treated as an `or` right?

carvantes: No. All filter clauses must match the document. i.e. it behaves like an `AND`

src/elasticSearchService.ts
carvantes: yes. 

In general most ES APIs do not make a distinction between an alias or an index

PR #85, opened 2021-06-28
feat: handle searches against static ES mapping

rsmayda:
Description of changes:
- Adding new constructor parameter to pass-in if customer is using static mapping or not
  - if `true` `.keyword` will be added to the restricted search queries

### DIFF
--- a/src/QueryBuilder/typeQueries/tokenQuery.test.ts
+++ b/src/QueryBuilder/typeQueries/tokenQuery.test.ts
@@ -2,7 +2,7 @@
  *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *  SPDX-License-Identifier: Apache-2.0
  */
-import each from 'jest-each';
+
 import { InvalidSearchParameterError } from 'fhir-works-on-aws-interface';
 import { parseTokenSearchParam, tokenQuery } from './tokenQuery';
 import { FHIRSearchParametersRegistry } from '../../FHIRSearchParametersRegistry';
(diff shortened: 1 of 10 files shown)


src/QueryBuilder/typeQueries/tokenQuery.test.ts
```diff
@@ -80,16 +80,17 @@ describe('parseTokenSearchParam', () => {
...
             Object {
               "bool": Object {
                 "must": Array [
                   Object {
                     "multi_match": Object {
                       "fields": Array [
-                        "identifier.system.keyword",
-                        "identifier.coding.system.keyword",
+                        "identifier.system${keywordSuffix}",
+                        "identifier.coding.system${keywordSuffix}",
```
carvantes: snapshots are plain text that's managed by jest. We shouldn't manually modify them. Also, dynamic content breaks them (the test pass, but jest can no longer update snapshots) 

In general `isESStaticallyTyped` has a minor effect on the ES query and I don't think we need to test it for ALL test cases. I suggest just adding an extra test with simple query params and `isESStaticallyTyped` set to true and leave the existing tests as is (with isESStaticallyTyped set to false)

src/elasticSearchService.ts
```diff
@@ -68,12 +71,14 @@ export class ElasticSearchService implements Search {
         fhirVersion: FhirVersion = '4.0.1',
         compiledImplementationGuides?: any,
         esClient: Client = ElasticSearch,
+        { isESStaticallyTyped = false }: { isESStaticallyTyped?: boolean } = {},
```
carvantes: nit: I'd rename it to something like `useKeywordSubFields`.

We know that this PR is being done because of switching away from dynamic mapping. But static typing on itself just means that you manually defined the mapping. You can manually create mappings that still use multi-fields with keyword type.

carvantes (changes requested):

src/QueryBuilder/typeQueries/tokenQuery.test.ts
rsmayda: Interesting I didn't know that was true when using the `toMatchInlineSnapshot` good to know

src/elasticSearchService.ts
```diff
@@ -59,6 +61,7 @@ export class ElasticSearchService implements Search {
      * @param compiledImplementationGuides - The output of ImplementationGuides.compile.
      * This parameter enables support for search parameters defined in Implementation Guides.
      * @param esClient
+     * @param options.useKeywordSubFields - whether or not you would want `.keyword` is appended to specifc search fields or not. You should use iff you do dynamic mapping
```
carvantes: typo: `iff`

carvantes:
```suggestion
     * @param options.useKeywordSubFields - whether or not to append `.keyword` to fields in search queries. You should enable this if you do dynamic mapping
```

carvantes (commented):
Overall changes look good. But unit tests are failing.

carvantes (approved):

PR #91, opened 2021-07-01
Allow SearchParams with just an expression or just an xpath.

justinusmenzel:
Description of changes:
We noticed that none of the 3 SearchParameters in http://hl7.org/fhir/us/davinci-drug-formulary/artifacts.html#2 are being compiled because they only use `expression` and are lacking an `xpath`. It seems having an `xpath` isn't really necessary since the compiler merges the results of the two expression parsers. 
So if only either `expression` or `xpath` are present we can just parse one expression and use its result. 

### DIFF
--- a/src/implementationGuides/__snapshots__/index.test.ts.snap
+++ b/src/implementationGuides/__snapshots__/index.test.ts.snap
@@ -168,6 +168,30 @@ Array [
 ]
 `;
 
+exports[`compile another where url value - MedicationKnowledge.extension.where(url='http://hl7.org/fhir/us/davinci-drug-formulary/StructureDefinition/usdf-PlanID-extension').value 1`] = `
+Array [
+  Object {
+    "base": "MedicationKnowledge",
+    "compiled": Array [
+      Object {
+        "condition": Array [
+          "extension.url",
+          "=",
+          "http://hl7.org/fhir/us/davinci-drug-formulary/StructureDefinition/usdf-PlanID-extension",
+        ],
+        "path": "extension.value",
+        "resourceType": "MedicationKnowledge",
+      },
+    ],
+    "description": "Accesses the DrugPlan ID of a FormularyDrug",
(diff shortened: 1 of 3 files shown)


src/implementationGuides/__snapshots__/index.test.ts.snap
```diff
@@ -189,6 +189,30 @@ Array [
...
+  Object {
+    "base": "MedicationKnowledge",
+    "compiled": Array [
+      Object {
+        "condition": Array [
+          "extension.url",
+          "=",
+          "http://hl7.org/fhir/us/davinci-drug-formulary/StructureDefinition/usdf-PlanID-extension",
+        ],
+        "path": "extension.value",
```
carvantes, line 203: The path is incorrect. It should be `extension.valueString` because it is defined as string on http://hl7.org/fhir/us/davinci-drug-formulary/STU1.0.1/StructureDefinition-usdf-PlanID-extension-definitions.html#Extension.value[x]

This is because `extension.value[x]` has choice of data types. See: https://github.com/awslabs/fhir-works-on-aws-search-es/pull/44

Here's an example of how Xpath is explicit about the types and the fhirpath expression is not (notice the `value` vs `valueCoding`):
https://github.com/awslabs/fhir-works-on-aws-search-es/blob/8de76c1056370e4964e0bef33ec161ef954ba2dd/src/implementationGuides/index.test.ts#L138-L141.

We are relying on the Xpath expression to resolve the choice of data types.

It would be possible to cross reference `SearchParameters` with the corresponding `StructureDefinition` to resolve the data types when the Xpath expression is not available, but that's no small task.

In the meanwhile it's worth considering manually adding the missing Xpath expressions with the resolved types on the IGs that don't have them. It's not ideal, but it can quickly unblock you.

justinusmenzel:
Closing this, since this obviously isn't the correct approach.

PR #93, opened 2021-07-12
feat: support OR search parameters (comma)

ssvegaraju:
Issue #, if available:

Description of changes:
- Modified search parameters to search on individual parameters delimited by a comma using a logical OR as specified in the [FHIR specification.]((https://www.hl7.org/fhir/search.html#composite)
- Search parameters separated by escaped commas will be treated as a single parameter.


src/QueryBuilder/index.ts
```diff
@@ -69,16 +69,37 @@ function typeQueryWithConditions(
...
+            if (c + 1 < searchValue.length && searchValue[c+1] === ',') {
+                // replace the escape character to allow the string to be handled by ES
+                searchValue = searchValue.substring(0, c) + searchValue.substring(c+1);
+            }
+        } else if (searchValue[c] === ',') {
+            splitSearchValue.push(searchValue.substring(lastIndex, c));
+            lastIndex = c + 1;
+        }
+    }
+    splitSearchValue.push(searchValue.substring(lastIndex));
```
carvantes: Move this to its own file and add unit tests. Similar to [sort.test.ts](https://github.com/awslabs/fhir-works-on-aws-search-es/blob/a7d9bf02228cf6d2b0efd5de608cd3ee4b5b3089/src/QueryBuilder/sort.test.ts#L12)

does this handle cases like `a\\,b` where it is not an escaped comma, but rather an escaped backslash?

carvantes (changes requested):
Things are looking good, but we need to add some more tests before merging.

Make sure to add a couple tests for the query as whole. You can add test parameters that use commas here:
https://github.com/awslabs/fhir-works-on-aws-search-es/blob/644e359d3e3409c726b788b159a5e8d26deb5b19/src/elasticSearchService.test.ts#L44

and verify that the snapshots look as expected

carvantes:
Make sure that all the automated GitHub checks pass when submitting a PR. Here are a few tips:

We use semantic commit messages as described on: https://www.conventionalcommits.org/en/v1.0.0/
In this case a good commit message would be `feat: support OR search parameters (comma)`. Rename the PR to have that title and also push an empty commit (`git commit --allow-empty -m "feat: support OR search parameters (comma)"`) with that message(the semantic pull request check only enforces that there's at least 1 commit with a semantic message)

You need to resolve the conflicts before this PR can be merged. pull the latest changes from mainline and fix the conflicts locally

I _think_ that the unit test and linting check will only run after the conflicts are fixed.

global-setup.js
```diff
@@ -0,0 +1,3 @@
+module.exports = () => {
+    process.env.TZ = 'UTC';
+}
```
carvantes: Did this actually fix the issue in windows?

src/__snapshots__/elasticSearchService.test.ts.snap
```diff
@@ -1443,40 +1443,46 @@ Array [
...
-                "multi_match": Object {
-                  "fields": Array [
-                    "id.code.keyword",
-                    "id.coding.code.keyword",
-                    "id.value.keyword",
-                    "id",
-                  ],
-                  "lenient": true,
-                  "query": "11111111-1111-1111-1111-111111111111",
+              Array [
```
carvantes: Nested arrays are not valid ES queries. Not sure why this is happening, but looks like the OR implementation is introducing a bug.

src/QueryBuilder/index.ts
```diff
@@ -73,16 +74,14 @@ function typeQueryWithConditions(
 }
 
 function searchParamQuery(searchParam: SearchParam, searchValue: string, useKeywordSubFields: boolean): any {
-    const queries = searchParam.compiled.map(compiled => {
-        return typeQueryWithConditions(searchParam, compiled, searchValue, useKeywordSubFields);
-    });
+    let queryList = getSearchQueries(searchValue, searchParam, useKeywordSubFields, typeQueryWithConditions);
```
carvantes: I think it would be better to have a simpler function like:
```ts
getOrSearchValues(searchValue: string): string[]
// i.e.  getOrSearchValues('aa,bb') returns ['aa', 'bb']
```

And then use those search values to assemble the queries here.

It's often better to avoid passing function as parameters(unless you are doing functional programming style function composition). It makes it harder to follow the code execution. Here passing `typeQueryWithConditions` as parameter is a bit weird since are likely never going to pass a different function as a parameter.

src/QueryBuilder/searchOR.ts
```diff
@@ -0,0 +1,35 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ */
+import { SearchParam } from '../FHIRSearchParametersRegistry';
+import { InvalidSearchParameterError } from 'fhir-works-on-aws-interface';
+
+export const getSearchQueries = (searchValue: string, searchParam: SearchParam, useKeywordSubFields: boolean,  typeQueryWithConditions: any) => {
```
carvantes: Avoid using the `any` type. It defeats the point of using Typescript. We do have several `any` in our codebase but we try not to. 
In this case `getOrSearchValues` is a function, so it's very important to declare the types to make sure the parameter is callable as a function and avoid runtime errors (although from my other comment on this PR it may be better to not pass function as parameters anyways)

carvantes (changes requested):
Good progress, but some changes are still needed.

Also, make sure that all automated checks pass on your PRs, this time there are several linting errors.
 
tip: run `yarn run release` before submitting a PR. It runs eslint and unit tests

Most lint errors can be fixed with `yarn run lint --fix`(your IDE may also have an integration with ESlint to highlight/fix linting errors)

global-setup.js
ssvegaraju: I thought it would, but it didn't actually end up fixing the issue on windows, so I removed it.

carvantes (commented):
Looks good.

Just one more thing before merging, can you add unit tests for `searchOR.ts`. Those tests are simpler since it's just an input string and compare it to an expected array of strings. We can cover more extreme edge cases in there like `Emily\\\\\\,Smith`

These unit tests may seem redundant with the ones `elasticSearchService.test.ts`. The tests`elasticSearchService.test.ts` are super useful, but they feel a bit like end to end tests and are a bit harder to read. We started with those since early on there were very few tests and the search code was a bit monolithic. As we continue to refactor and split the logic into smaller functions we want to add unit tests for the smaller parts.

carvantes (approved):

PR #96, opened 2021-07-28
chore(release): 3.2.0

ssvegaraju:
Issue #, if available:

Description of changes:


package.json
```diff
@@ -53,6 +53,7 @@
     "jest-mock-extended": "^1.0.8",
     "json-stringify-pretty-compact": "^2.0.0",
     "prettier": "^1.19.1",
+    "standard-version": "^9.3.1",
```
carvantes: This dependency shouldn’t be here. It should only be installed globally on your mac/pc

Also delete the package-lock.json file

in general release PRs only bump the version and update the changelog

carvantes (changes requested):

package.json
ssvegaraju: Just removed the dependency on standard-version and removed the package-lock.json file.

carvantes (approved):

PR #98, opened 2021-08-02
fix: remove duplicate Resource search params

carvantes:
Fixes part of https://github.com/awslabs/fhir-works-on-aws-routing/issues/106

The capability statement validation was failing with errors like:
`ERROR: Search parameter names must be unique in the context of a resource. [searchParam.select(name).isDistinct()]`

Search params for `Resource` are applied to all resources (e.g. the `_id` param). The bug was that this params were added without checking if there was already a more specific param with the same name for a specific resource type(e.g. US Core defines a more specific `_id` for `Encounter`)

### DIFF
--- a/src/FHIRSearchParametersRegistry/index.ts
+++ b/src/FHIRSearchParametersRegistry/index.ts
@@ -148,6 +148,8 @@ export class FHIRSearchParametersRegistry {
 
         const resourceSearchParams = Object.values(this.typeNameMap.Resource).map(toCapabilityStatement);
 
+        // For each resource type, add all search params that have "Resource" as base, except when there is already
+        // a more specific search parameter with the same name.
         Object.entries(this.capabilityStatement).forEach(([resourceType, searchCapabilities]) => {
             searchCapabilities.searchParam.push(
                 ...resourceSearchParams.filter(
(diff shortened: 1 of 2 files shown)


src/FHIRSearchParametersRegistry/index.ts
```diff
@@ -148,8 +148,12 @@ export class FHIRSearchParametersRegistry {
 
         const resourceSearchParams = Object.values(this.typeNameMap.Resource).map(toCapabilityStatement);
 
-        Object.values(this.capabilityStatement).forEach(searchCapabilities => {
-            searchCapabilities.searchParam.push(...resourceSearchParams);
+        Object.entries(this.capabilityStatement).forEach(([resourceType, searchCapabilities]) => {
+            searchCapabilities.searchParam.push(
+                ...resourceSearchParams.filter(
+                    resourceSearchParam => !this.typeNameMap?.[resourceType]?.[resourceSearchParam.name],
```
rsmayda, line 156: Could you add a comment, explaining the loop

"For each resource, add all parameters that ..."

carvantes: done

rsmayda (approved):

Issue #106, opened 2021-08-20
[Feature Request] support both [type]/[id] and [id] when searching by reference


ghost:
**Is your feature request related to a problem? Please describe.**
Inferno search by patient tests fail

![image](https://user-images.githubusercontent.com/69013924/130281735-02de1c6d-856e-4d41-965c-4e7d124184bd.png)

**Describe the solution you'd like**
Both forms of the search to return the same results

**Describe alternatives you've considered**

**Additional context**
US Core search guidance states servers shall support both forms of the reference search

http://hl7.org/fhir/us/core/general-guidance.html#search-syntax

rsmayda:
Thanks for bringing this up we will add it to our backlog and prioritize it soon!

rsmayda:
FYI in situations where there is multiple types allowed, this is what we will do:

> Some references may point to more than one type of resource; e.g. subject: Reference(Patient|Group|Device|..). In these cases, multiple resources may have the same logical identifier. Servers SHOULD reject a search where the logical id refers to more than one matching resource across different types. In order to allow the client to perform a search in these situations the type is specified explicitly

http://hl7.org/fhir/R4/search.html#reference

carvantes:
> FYI in situations where there is multiple types allowed, this is what we will do:
> 
> > Some references may point to more than one type of resource; e.g. subject: Reference(Patient|Group|Device|..). In these cases, multiple resources may have the same logical identifier. Servers SHOULD reject a search where the logical id refers to more than one matching resource across different types. In order to allow the client to perform a search in these situations the type is specified explicitly
> 
> http://hl7.org/fhir/R4/search.html#reference

@rsmayda rejecting such requests is a SHOULD not a SHALL, so it's not hard requirement. I do think it's easier to simply reject them, but some FHIR servers accept them

Notably, HAPI actually accepts such requests. All the following queries are valid even though the `subject` parameter has an ambiguous target resource type (`Observation.subject(Group, Device, Patient, Location)`)

http://hapi.fhir.org/baseR4/Observation?subject=247816
http://hapi.fhir.org/baseR4/Observation?subject=Group/247816
http://hapi.fhir.org/baseR4/Observation?subject=Patient/126
http://hapi.fhir.org/baseR4/Observation?subject=126

rsmayda:
@medhost-bfindeisen is the rejection behavior okay? 
[...]

PR #107, opened 2021-08-25
feat: allow id only reference searches when valid

rsmayda:
Issue #, if available: #106

FHIR spec:
> Some references may point to more than one type of resource; e.g. subject: Reference(Patient|Group|Device|..). In these cases, multiple resources may have the same logical identifier. Servers SHOULD reject a search where the logical id refers to more than one matching resource across different types. In order to allow the client to perform a search in these situations the type is specified explicitly

http://hl7.org/fhir/R4/search.html#reference

### DIFF
--- a/src/QueryBuilder/typeQueries/referenceQuery.test.ts
+++ b/src/QueryBuilder/typeQueries/referenceQuery.test.ts
@@ -14,12 +14,10 @@ describe('referenceQuery', () => {
     test('simple value; with keyword', () => {
         expect(referenceQuery(organizationParam, 'Organization/111', true, 'organization', [])).toMatchInlineSnapshot(`
             Object {
-              "multi_match": Object {
-                "fields": Array [
-                  "managingOrganization.reference.keyword",
+              "terms": Object {
+                "managingOrganization.reference.keyword": Array [
+                  "Organization/111",
                 ],
-                "lenient": true,
-                "query": "Organization/111",
               },
             }
         `);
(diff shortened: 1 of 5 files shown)


ssvegaraju (approved):

src/QueryBuilder/typeQueries/referenceQuery.test.ts
```diff
@@ -24,16 +25,41 @@ describe('referenceQuery', () => {
...
                 ],
                 "lenient": true,
-                "query": "Organization/111",
+                "query": "http://fhir.com/baseR4/Organization/111",
               },
             }
         `);
     });
+    test('just id search, one type found', () => {
+        expect(referenceQuery(organizationParam, 'orgnaizationId', false, 'organization', ['Organization']))
```
carvantes: typo `orgnaizationId`

src/QueryBuilder/typeQueries/referenceQuery.test.ts
```diff
@@ -24,16 +25,41 @@ describe('referenceQuery', () => {
...
+                ],
+                "lenient": true,
+                "query": "Organization/orgnaizationId",
+              },
+            }
+        `);
+    });
+    test('just id search, many types found', () => {
+        expect(() =>
+            referenceQuery(organizationParam, 'orgnaizationId', false, 'organization', ['Organization', 'Group']),
```
carvantes: typo `orgnaizationId`

src/QueryBuilder/typeQueries/referenceQuery.test.ts
```diff
@@ -24,16 +25,41 @@ describe('referenceQuery', () => {
...
+        ).toThrow(InvalidSearchParameterError);
+    });
+    test('just id search, no types found', () => {
+        expect(() => referenceQuery(organizationParam, 'orgnaizationId', false, 'organization')).toThrow(
```
carvantes: typo `orgnaizationId`

src/QueryBuilder/typeQueries/referenceQuery.ts
```diff
@@ -3,17 +3,39 @@
...
     const fields = [`${compiled.path}.reference${keywordSuffix}`];
+    let reference = value;
+    // http://hl7.org/fhir/R4/search.html#reference
+    // to cover the use-case of someone searching as just an 'id' and this reference just has 1 target i.e. Observation.patient = 123
+    if (value.match(idOnlyRegExp)) {
+        if (target.length === 1) {
+            reference = `${target[0]}/${value}`;
+        } else {
+            throw new InvalidSearchParameterError(
+                `'${searchParamName}' is invalid please specify the resource type with the id`,
```
carvantes: I think that the error message is hard to understand if you don't know the FHIR spec beforehand.
Maybe it's clearer w something like
```
`'${searchParamName}' is invalid please specify the reference value with the format [resourcetType]/[id] or as an absolute URL`

```

carvantes (approved):
looks good. 

Only minor comments related to typos and text messages.

src/QueryBuilder/typeQueries/referenceQuery.ts
```diff
@@ -3,18 +3,35 @@
...
-    };
+    // http://hl7.org/fhir/R4/search.html#reference
+    // reference fields should be of `keyword` ES type: https://www.elastic.co/guide/en/elasticsearch/reference/current/keyword.html
+    let references: string[] = [value];
+    if (value.match(idOnlyRegExp)) {
+        if (target.length === 0) {
+            throw new InvalidSearchParameterError(
+                `'${searchParamName}' is invalid please specify the reference value with the format [resourcetType]/[id] or as an absolute URL`,
+            );
+        }
```
carvantes: I think this is no longer the correct error msg. 
Looks like if `target.length === 0` then there's something wrong with the conpiler (or the FHIR definitions)

src/QueryBuilder/typeQueries/referenceQuery.test.ts
```diff
@@ -11,29 +12,56 @@ const organizationParam = fhirSearchParametersRegistry.getSearchParameter('Patie
...
                 ],
-                "lenient": true,
-                "query": "Organization/111",
               },
             }
         `);
     });
     test('simple value; without keyword', () => {
-        expect(referenceQuery(organizationParam, 'Organization/111', false)).toMatchInlineSnapshot(`
+        expect(referenceQuery(organizationParam, 'http://fhir.com/baseR4/Organization/111', false, 'organization'))
```
carvantes: can we move this to a separate test? something like "search for full URL".
I think that it's an important case that deserves it's own test.

carvantes (changes requested):

src/QueryBuilder/typeQueries/referenceQuery.ts
rsmayda: Agreed. I am in favor of still throwing a `InvalidSearchParameterError` instead of a `InternalServerError`, to unblock the requester and give them advice on how to continue searching. While logging an error for the system admins

carvantes: I agree. 

It'll be a bit weird for the customer if the search query was correct, but FWoA is missing the target values. But that should be a very rare case anyways.

We should also at some point add validation on the compiler to ensure that reference parameters have at least one target.

carvantes (approved):

PR #108, opened 2021-08-26
feat: add support for exact search modifier

ssvegaraju:
- feat: add support for exact search modifier
- feat: add implementation of 'exact' search modifier

Issue #, if available:

Description of changes:

### DIFF
--- a/src/QueryBuilder/index.ts
+++ b/src/QueryBuilder/index.ts
@@ -3,7 +3,7 @@
  *  SPDX-License-Identifier: Apache-2.0
  */
 
-import { InvalidSearchParameterError, ResourceConflictError, TypeSearchRequest } from 'fhir-works-on-aws-interface';
+import { InvalidSearchParameterError, TypeSearchRequest } from 'fhir-works-on-aws-interface';
 import { NON_SEARCHABLE_PARAMETERS } from '../constants';
 import { CompiledSearchParam, FHIRSearchParametersRegistry, SearchParam } from '../FHIRSearchParametersRegistry';
 import { stringQuery } from './typeQueries/stringQuery';
@@ -126,7 +126,7 @@ function searchRequestQuery(
     return Object.entries(normalizeQueryParams(queryParams))
         .filter(([searchParameter]) => !NON_SEARCHABLE_PARAMETERS.includes(searchParameter))
         .flatMap(([searchParameter, searchValues]) => {
-            const unmodifiedParam = searchParameter.split(":")[0];
+            const unmodifiedParam = searchParameter.split(':')[0];
             const fhirSearchParam = fhirSearchParametersRegistry.getSearchParameter(resourceType, unmodifiedParam);
             if (fhirSearchParam === undefined) {
                 throw new InvalidSearchParameterError(
@@ -138,8 +138,9 @@ function searchRequestQuery(
             if (searchModifier === SearchModifier.Exact) {
                 forceUseKeywordSubFields = true;
             }
-            return searchValues.map(searchValue => searchParamQuery(fhirSearchParam, searchValue,
-                forceUseKeywordSubFields? true: useKeywordSubFields));
+            return searchValues.map(searchValue =>
+                searchParamQuery(fhirSearchParam, searchValue, forceUseKeywordSubFields ? true : useKeywordSubFields),
+            );
         });
 }
 
--- a/src/QueryBuilder/index.ts
+++ b/src/QueryBuilder/index.ts
@@ -13,7 +13,7 @@ import { numberQuery } from './typeQueries/numberQuery';
 import { quantityQuery } from './typeQueries/quantityQuery';
 import { referenceQuery } from './typeQueries/referenceQuery';
 import getOrSearchValues from './searchOR';
-import getSearchModifiers, { SearchModifier } from './searchModifiers';
+import parseSearchModifiers from './searchModifiers';
 import { uriQuery } from './typeQueries/uriQuery';
 
 function typeQueryWithConditions(
@@ -21,34 +21,35 @@ function typeQueryWithConditions(
     compiledSearchParam: CompiledSearchParam,
     searchValue: string,
     useKeywordSubFields: boolean,
+    modifier?: string,
 ): any {
     let typeQuery: any;
     switch (searchParam.type) {
         case 'string':
-            typeQuery = stringQuery(compiledSearchParam, searchValue, useKeywordSubFields);
+            typeQuery = stringQuery(compiledSearchParam, searchValue, modifier);
             break;
         case 'date':
-            typeQuery = dateQuery(compiledSearchParam, searchValue);
+            typeQuery = dateQuery(compiledSearchParam, searchValue, modifier);
             break;
         case 'token':
-            typeQuery = tokenQuery(compiledSearchParam, searchValue, useKeywordSubFields);
+            typeQuery = tokenQuery(compiledSearchParam, searchValue, useKeywordSubFields, modifier);
             break;
         case 'number':
-            typeQuery = numberQuery(compiledSearchParam, searchValue);
+            typeQuery = numberQuery(compiledSearchParam, searchValue, modifier);
             break;
         case 'quantity':
-            typeQuery = quantityQuery(compiledSearchParam, searchValue, useKeywordSubFields);
+            typeQuery = quantityQuery(compiledSearchParam, searchValue, useKeywordSubFields, modifier);
             break;
         case 'reference':
-            typeQuery = referenceQuery(compiledSearchParam, searchValue, useKeywordSubFields);
+            typeQuery = referenceQuery(compiledSearchParam, searchValue, useKeywordSubFields, modifier);
             break;
         case 'uri':
-            typeQuery = uriQuery(compiledSearchParam, searchValue, useKeywordSubFields);
+            typeQuery = uriQuery(compiledSearchParam, searchValue, useKeywordSubFields, modifier);
             break;
         case 'composite':
         case 'special':
         default:
-            typeQuery = stringQuery(compiledSearchParam, searchValue, useKeywordSubFields);
+            typeQuery = stringQuery(compiledSearchParam, searchValue, modifier);
     }
     // In most cases conditions are used for fields that are an array of objects
     // Ideally we should be using a nested query, but that'd require to update the index mappings.
@@ -74,13 +75,24 @@ function typeQueryWithConditions(
     return typeQuery;
 }
 
-function searchParamQuery(searchParam: SearchParam, searchValue: string, useKeywordSubFields: boolean): any {
+function searchParamQuery(
+    searchParam: SearchParam,
+    searchValue: string,
+    useKeywordSubFields: boolean,
+    modifier?: string,
+): any {
     const splitSearchValue = getOrSearchValues(searchValue);
     let queryList = [];
     for (let i = 0; i < splitSearchValue.length; i += 1) {
         queryList.push(
             searchParam.compiled.map(compiled => {
-                return typeQueryWithConditions(searchParam, compiled, splitSearchValue[i], useKeywordSubFields);
+                return typeQueryWithConditions(
+                    searchParam,
+                    compiled,
+                    splitSearchValue[i],
+                    useKeywordSubFields,
+                    modifier,
+                );
             }),
         );
     }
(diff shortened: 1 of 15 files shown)


src/QueryBuilder/typeQueries/stringQuery.ts
```diff
@@ -10,8 +10,10 @@ const escapeQueryString = (string: string) => {
 };
 
 // eslint-disable-next-line import/prefer-default-export
-export function stringQuery(compiled: CompiledSearchParam, value: string): any {
-    const fields = [compiled.path, `${compiled.path}.*`];
+export function stringQuery(compiled: CompiledSearchParam, value: string, useKeywordSubFields: boolean): any {
+    const keywordSuffix = useKeywordSubFields ? '.keyword' : '';
+    const fields = [compiled.path, `${compiled.path}.*${keywordSuffix}`];
```
carvantes: I think that this won't work as expected. for `:exact` we want to query exclusively `keyword` type fields. `compiled.path` in many cases is a `text` field

src/elasticSearchService.ts
```diff
@@ -185,7 +185,6 @@ export class ElasticSearchService implements Search {
                     request.queryParams[SORT_PARAMETER],
                 );
             }
-
```
carvantes, line 188: nit: try to avoid this format changes. They just make the PRs bigger for no good reason

src/QueryBuilder/searchModifiers.ts
```diff
@@ -0,0 +1,27 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ */
+
+export enum SearchModifier {
+    None = 'none',
+    Error = 'error',
```
carvantes: Where is this used? It's probably better to throw an exception instead

src/QueryBuilder/index.ts
```diff
@@ -125,13 +126,21 @@ function searchRequestQuery(
     return Object.entries(normalizeQueryParams(queryParams))
         .filter(([searchParameter]) => !NON_SEARCHABLE_PARAMETERS.includes(searchParameter))
         .flatMap(([searchParameter, searchValues]) => {
-            const fhirSearchParam = fhirSearchParametersRegistry.getSearchParameter(resourceType, searchParameter);
+            const unmodifiedParam = searchParameter.split(':')[0];
```
carvantes: Modifiers parsing logic (such as `.split(':')[0];`) should be defined in a single place. Probably on `searchModifiers.ts`

src/QueryBuilder/index.ts
```diff
@@ -125,13 +126,21 @@ function searchRequestQuery(
...
-                    `Invalid search parameter '${searchParameter}' for resource type ${resourceType}`,
+                    `Invalid search parameter '${unmodifiedParam}' for resource type ${resourceType}`,
                 );
             }
-            return searchValues.map(searchValue => searchParamQuery(fhirSearchParam, searchValue, useKeywordSubFields));
+            const searchModifier = getSearchModifiers(searchParameter);
+            let forceUseKeywordSubFields = false;
+            if (searchModifier === SearchModifier.Exact) {
+                forceUseKeywordSubFields = true;
+            }
```
carvantes: I don't think we should conflate the `:exact` modifier with the `useKeywordSubFields` param.

We need to validate that `:exact` is used ONLY with search params of type `string`. Only `stringQuery` needs to know how to implement `:exact`(which is about using `keyword` fields)., whereas `useKeywordSubFields` is a more generic param that several query types use

src/QueryBuilder/searchModifiers.ts
```diff
@@ -0,0 +1,27 @@
...
+    Exact = 'exact',
+}
+
+const getSearchModifiers = (searchParameter: string): SearchModifier => {
```
carvantes: consider having this function return the param name as well
```ts
{
  searchParameterName: string,
  modifier?: string,
}
```

also nit: maybe rename to `parseSearchModifiers`

src/QueryBuilder/typeQueries/stringQuery.ts
```diff
@@ -10,8 +10,10 @@ const escapeQueryString = (string: string) => {
 };
 
 // eslint-disable-next-line import/prefer-default-export
-export function stringQuery(compiled: CompiledSearchParam, value: string): any {
-    const fields = [compiled.path, `${compiled.path}.*`];
+export function stringQuery(compiled: CompiledSearchParam, value: string, useKeywordSubFields: boolean): any {
```
carvantes: I think that we should update `stringQuery` (and all the other `somethingQuery`) to take a `modifier` parameter.

Then each method can be responsible of ensuring that the modifier is applicable to the search param type and of implementing it.

carvantes (changes requested):

carvantes:
btw, rename the PR to be semantic. Note the extra `- ` before `feat`.

In theory the only thing that matters is the commit message when doing squash and merge, but it defaults to the PR name and it's easy to miss updating it later

src/__snapshots__/elasticSearchService.test.ts.snap
```diff
@@ -1652,7 +1652,7 @@ Array [
                 "multi_match": Object {
                   "fields": Array [
                     "name",
-                    "name.*",
+                    "name.*.keyword",
```
carvantes: Maybe I'm missing something, but looks like all the extra `.keyword` on snapshots are actually breaking the tests.  The search query is not using `:exact` so it shouldn't append `.keyword`

src/QueryBuilder/typeQueries/stringQuery.ts
ssvegaraju: I switched where we use the keyword subfield and it appears to work as expected, for example, in failing to find results on case mismatch.

src/elasticSearchService.ts
ssvegaraju: Got it, will keep that in mind going forward!

src/QueryBuilder/searchModifiers.ts
ssvegaraju: I replaced this enum with the return object you suggested in a different message. I initially thought it may be better to allow callers to handle errors but it doesn't quite make sense for them to handle parsing errors.

src/QueryBuilder/index.ts
ssvegaraju: All of the parsing logic is now encapsulated in `parseSearchModifiers`

src/QueryBuilder/index.ts
ssvegaraju: This makes complete sense, I confused myself into thinking keyword sub-fields involved this parameter. This has been changed to simply use a new parameter instead.

src/QueryBuilder/searchModifiers.ts
ssvegaraju: changed the return type and renamed the function!

src/QueryBuilder/typeQueries/stringQuery.ts
ssvegaraju: I've updated all the query types to ensure that the given modifier is valid for the type of query (in this case, exact is only applicable to string queries).

src/__snapshots__/elasticSearchService.test.ts.snap
ssvegaraju: I restored the previously working snapshot in the latest commit, and made sure that all other tests pass even without a specified modifier and don't add anything to the output.

src/QueryBuilder/typeQueries/stringQuery.ts
```diff
@@ -10,8 +10,10 @@ const escapeQueryString = (string: string) => {
 };
 
 // eslint-disable-next-line import/prefer-default-export
-export function stringQuery(compiled: CompiledSearchParam, value: string): any {
-    const fields = [compiled.path, `${compiled.path}.*`];
+export function stringQuery(compiled: CompiledSearchParam, value: string, modifier?: string): any {
+    const keywordSuffix = modifier === 'exact' ? '.keyword' : '';
+    const fields = [compiled.path + keywordSuffix, `${compiled.path}.*`];
+
```
carvantes: We should throw an error if the modifier is something other than `exact`

src/QueryBuilder/typeQueries/dateQuery.ts
```diff
@@ -68,7 +68,10 @@ export const parseDateSearchParam = (param: string): DateSearchParameter => {
 };
 
 // eslint-disable-next-line import/prefer-default-export
-export const dateQuery = (compiledSearchParam: CompiledSearchParam, value: string): any => {
+export const dateQuery = (compiledSearchParam: CompiledSearchParam, value: string, modifier?: string): any => {
+    if (modifier === 'exact') {
+        throw new InvalidSearchParameterError(`Invalid date search modifier: ${modifier}`);
+    }
```
carvantes: We should "flip" this validation logic. Right now we'd need to update the validation rules in ALL queries when we add a new modifier. It would be simpler to have something like `const SUPPORTED_MODIFIERS = []` and use it for validation. 
Then for new modifiers we just need to modify the queries that actually support them.
Same goes for all other typeQueries files

src/QueryBuilder/searchModifiers.ts
```diff
@@ -0,0 +1,29 @@
...
+    if (modifier.length === 1) {
+        return { parameterName: modifier[0], modifier: undefined };
+    }
+    modifier[1] = modifier[1].toLowerCase();
+    switch (modifier[1]) {
+        case 'exact':
+            return { parameterName: modifier[0], modifier: modifier[1] };
+        default:
+            throw new InvalidSearchParameterError(`Unsupported search modifier: ${modifier[1]}`);
+    }
```
carvantes: I think that we can remove this validation. The thing is that the full validation requires knowing the type of the search parameter to check if the modifier is applicable to that type and that is done on the typeQueries. This partial validation feels redundant.

carvantes (changes requested):

src/QueryBuilder/typeQueries/stringQuery.ts
ssvegaraju: Good catch, I updated this as well as all the other queries to utilize the `SUPPORTED_MODIFIERS` `const` to check if a modifier is valid.

src/QueryBuilder/searchModifiers.ts
ssvegaraju: That makes sense to me, I've gone ahead and removed it.

src/QueryBuilder/typeQueries/dateQuery.ts
ssvegaraju: Redid the validation logic according to this suggestion, makes much more sense than the branching if-else that I had previously.

src/QueryBuilder/typeQueries/stringQuery.test.ts
```diff
@@ -66,4 +66,32 @@ describe('stringQuery', () => {
...
+          }
+      `);
+    });
+    test('simple value with exact modifier and case differences', () => {
+        expect(stringQuery(nameParam, 'RoBeRt BeLL', 'exact')).toMatchInlineSnapshot(`
+          Object {
+            "multi_match": Object {
+              "fields": Array [
+                "name.keyword",
+                "name.*",
```
carvantes: I think this is not working as intended, it should query `name.*.keyword`, otherwise it may include text fields

This was correct on a previous version. Related to https://github.com/awslabs/fhir-works-on-aws-search-es/pull/108#discussion_r696906778

carvantes (changes requested):

src/QueryBuilder/typeQueries/stringQuery.test.ts
ssvegaraju: I see, I think I misunderstood the comment but I've restored the correct functionality in the latest commit.

carvantes: Last commit is not correct yet. **Both** fields must end in `.keyword`. (`"name.keyword"`, `"name.*.keyword"`)

ssvegaraju: Got it, fixed that in the latest commit

carvantes (approved):
🚀🚀🚀

PR #113, opened 2021-09-02
feat: implement ":contains" modifier

carvantes:
> [base]/Patient?given:contains=eve - Any  patients with a name with a given part containing "eve" at any  position. This would include patients with the given name "Eve",  "Evelyn", and also "Severine".

https://www.hl7.org/fhir/search.html#string


README.md
```diff
@@ -50,7 +50,7 @@ This package is dependent on:
 |---------------------|:----------:|-------|
 | `:missing`            |      -     |       |
 | `:exact`              |      -     |       |
```
ssvegaraju: this is a very small thing, but I didn't actually update the readme when implementing `exact:`, so maybe we could update that here?

ssvegaraju (approved):

src/QueryBuilder/typeQueries/stringQuery.ts
```diff
@@ -10,13 +10,55 @@ const escapeQueryString = (string: string) => {
...
         throw new InvalidSearchParameterError(`Unsupported string search modifier: ${modifier}`);
     }
+
+    if (modifier === 'contains') {
+        const fields = [`${compiled.path}`];
+
+        if (compiled.path === 'name') {
+            // name is a special parameter.
+            // name.* cannot be used in wildcard queries so we specify the fields to search.
+            fields.push('name.family', 'name.given', 'name.text', 'name.prefix', 'name.suffix');
```
rsmayda, line 27: Can we pull this from the IG compiler at all? (same for addresses)

carvantes: Not really, the name param just has a text description that reads: 

> A  server defined search that may match any of the string fields in the  HumanName, including family, give, prefix, suffix, suffix, and/or text

README.md
carvantes: yeah, I just updated it

rsmayda (approved):

PR #115, opened 2021-09-07
 fix: further reference support

rsmayda:
Description of changes:
Expand the support for references so that it behaves as the FHIR spec suggests: http://hl7.org/fhir/R4/search.html#reference
- When searching for just an id it should also match the id AND resourceType+ id AND fhir url + resourcetype + id
- When searching for resourceType + id it should also match fhir url + resourcetype + id
- When searching for fhir url + resourcetype + id  it should also match resourcetype + id IF fhir url === baseUrl


src/QueryBuilder/typeQueries/referenceQuery.test.ts
```diff
@@ -11,68 +11,145 @@ const fhirSearchParametersRegistry = new FHIRSearchParametersRegistry('4.0.1');
...
+    test('invalid modifier', () => {
+        expect(() =>
+            referenceQuery(
+                organizationParam,
+                'organizationId',
+                false,
+                'https://base-url.com',
+                'organization',
+                [],
+                'exact',
```
carvantes: I think that this empty array param makes this scenario overlap with the `no target types found`. We should pass a valid array to make sure that the test fails because of the modifier validation.

src/QueryBuilder/typeQueries/referenceQuery.test.ts
```diff
@@ -11,68 +11,145 @@ const fhirSearchParametersRegistry = new FHIRSearchParametersRegistry('4.0.1');
...
+            ),
+        ).toThrow(InvalidSearchParameterError);
     });
-    test('just id search, no types found', () => {
-        expect(() => referenceQuery(organizationParam, 'organizationId', false, 'organization')).toThrow(
-            InvalidSearchParameterError,
-        );
+    test('search does not match expect id or url', () => {
```
carvantes: nit: I think the test name could be a bit more descriptive
```suggestion
    test('search value is not an URL nor has the format <resourceType>/<id>', () => {
```

src/QueryBuilder/typeQueries/referenceQuery.ts
```diff
@@ -9,7 +9,8 @@ import getComponentLogger from '../../loggerBuilder';
 
 const logger = getComponentLogger();
 
-const idOnlyRegExp = /^[A-Za-z0-9\-.]{1,64}$/;
+const ID_ONLY_REGEX = /^[A-Za-z0-9\-.]{1,64}$/;
+const FHIR_RESOURCE_REGEX = /^((?<hostname>https?:\/\/[A-Za-z0-9\-\\.:%$_/]+)\/)?(?<resourceType>[A-Z][a-zA-Z]+)\/(?<id>[A-Za-z0-9\-.]{1,64})$/;
```
carvantes: Can we rename `hostname` to something like `fhirServiceBaseUrl`? I started switching to it on some parts of the interface: https://github.com/awslabs/fhir-works-on-aws-interface/blob/c46346932d71b0ee5e6c52a952c76f1c1123436c/src/authorization.ts#L46-L50

hostname is misleading. In the URL `https://fwoa.com/tenant/tenant1/Patient/111`, the actual hostname is `fwoa.com` but our hostname regex matches `https://fwoa.com/tenant/tenant1`. This discrepancy can be especially confusing when working with multi-tenant URLs.

carvantes: nit: maybe rename to `FHIR_REFERENCE_REGEX`

src/QueryBuilder/typeQueries/referenceQuery.test.ts
rsmayda: Excellent eye thank you

carvantes (approved):

PR #119, opened 2021-09-15
fix: fix _id searches

rsmayda:
Description of changes:
Due to [previous change](https://github.com/awslabs/fhir-works-on-aws-search-es/pull/117) _id searches are not working since by default we [map `id` to be a keyword](https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/blob/f9c9414a326f2e458b4e1c82ad29a694facee70e/src/ddbToEs/ddbToEsHelper.ts#L118-L119)

This change takes this into account and handles `id` as an exceptional case

### DIFF
--- a/src/QueryBuilder/typeQueries/tokenQuery.ts
+++ b/src/QueryBuilder/typeQueries/tokenQuery.ts
@@ -14,7 +14,7 @@ interface TokenSearchParameter {
 
 const SUPPORTED_MODIFIERS: string[] = [];
 
-// Fields that do not have `.keyword` suffix; currently it is just `id`. This is only important if `useKeywordSubFields` is true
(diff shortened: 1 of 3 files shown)


src/QueryBuilder/typeQueries/tokenQuery.ts
```diff
@@ -14,6 +14,9 @@ interface TokenSearchParameter {
 
 const SUPPORTED_MODIFIERS: string[] = [];
 
+// Fields that do not have `.keyword` suffix. This is only important if `useKeywordSubFields` is true
+const FIELDS_WITHOUT_KEYWORD = ['id'];
```
carvantes, line 18: nit: maybe rename to `FIELDS_WITHOUT_KEYWORD_SUBFIELD`

carvantes (approved):

PR #123, opened 2021-09-22
fix: update IG validation to reject reference params with no target

ssvegaraju:
Issue #, if available:
[#449](https://github.com/awslabs/fhir-works-on-aws-deployment/issues/449)
Description of changes:

### DIFF
--- a/src/implementationGuides/index.test.ts
+++ b/src/implementationGuides/index.test.ts
@@ -237,4 +237,22 @@ describe('compile', () => {
         ]);
         await expect(compiled).rejects.toThrowError();
     });
+
+    test(`search param of type reference with empty target`, async () => {
+        const compiled = compile([
+            {
+                resourceType: 'SearchParameter',
+                url: 'http://hl7.org/fhir/SearchParameter/ConceptMap-source-uri',
+                name: 'source-uri',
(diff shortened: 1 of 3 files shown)


src/implementationGuides/index.test.ts
```diff
@@ -220,4 +221,20 @@ describe('compile', () => {
         ]);
         await expect(compiled).rejects.toThrowError();
     });
+    test(`search param of type reference with no target`, async () => {
```
carvantes, line 224: Add a test for target being an empty array. I think a code change is needed to handle that case properly

carvantes (changes requested):

carvantes:
Also, the error msg of `Error: The following input is not a search parameter: {...}` is not that helpful. The `isFhirSearchParam` is a very basic check to make sure that required fields are present and that the field types are correct. 

Maybe we can create a separate method for semantic validations like this that throws more specific error messages.

src/implementationGuides/index.test.ts
ssvegaraju: Good catch, I added a test case for this!

carvantes (approved):

PR #124, opened 2021-09-24
feat: add static search mappings

carvantes:
Create static search mappings for all FHIR resource types.

There are 2 parts:

- `scripts/elasticSearchMappingsGenerator`: A script that generates the `src/schema/searchMappingsBase` files. This script parses the FHIR definitions and extracts the types for all the fields that are used on search parameters.
It is similar in use to the script that generates the compiled search params.

- `src/searchMappings`: A function that exposes the search mappings to customers. It takes the `src/schema/searchMappingsBase` files and adds the mappings for the field types to produce the final mappings object. Later on this function will be called to create search indices when deploying FWoA.


ssvegaraju (approved):
Looks good to me! Just one thing: is it worth it to add more tests in `index.test.ts` to cover FHIR versions other than 4.0.1 to ensure that the mappings are still valid? And should any of the smaller methods have unit tests of their own (such as `elasticSearchMappingsGenerator/index.ts`)?

carvantes:
> Looks good to me! Just one thing: is it worth it to add more tests in `index.test.ts` to cover FHIR versions other than 4.0.1 to ensure that the mappings are still valid? 

In `searchMappings/index.ts` there isn't any significant difference between FHIR versions other than the`searchMappingsBase` file being used. Adding a test for FHIR 3.0.1 is not that useful, and it results in a much larger snapshot file (it is already 2.5MB).

> And should any of the smaller methods have unit tests of their own (such as `elasticSearchMappingsGenerator/index.ts`)?

The methods under `elasticSearchMappingsGenerator/` are meant to run only once and then commit the results. In a sense, the `searchMappingsBase.x.0.1` file serves as a snapshot test for `elasticSearchMappingsGenerator/`

PR #126, opened 2021-10-07
feat: add class to aid in updating search mappings

carvantes:
Add `SearchMappingsManager` class that will be used by a custom resource to update the search mappings for all FHIR resource types.

Related PR:
- https://github.com/awslabs/fhir-works-on-aws-deployment/pull/474

### DIFF
--- a/src/searchMappingsManager/index.ts
+++ b/src/searchMappingsManager/index.ts
@@ -1,0 +1,77 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ */
+
+import { ResponseError } from '@elastic/elasticsearch/lib/errors';
+import { Client } from '@elastic/elasticsearch';
+import { ElasticSearch } from '../elasticSearch';
+
+const toIndexName = (resourceType: string) => resourceType.toLowerCase();
+
+// eslint-disable-next-line import/prefer-default-export
+export class SearchMappingsManager {
+    private readonly searchMappings: {
+        [resourceType: string]: any;
+    };
+
+    private readonly numberOfShards: number;
+
+    private readonly searchClient: Client;
+
+    constructor({
+        searchMappings,
+        numberOfShards,
+        searchClient = ElasticSearch,
+    }: {
+        searchMappings: { [resourceType: string]: any };
+        numberOfShards: number;
+        searchClient?: Client;
+    }) {
+        this.searchMappings = searchMappings;
+        this.numberOfShards = numberOfShards;
+        this.searchClient = searchClient;
+    }
+
+    /**
+     * Updates the mappings for all the FHIR resource types. If an index does not exist, it is created
+     */
+    async createOrUpdateMappings() {
+        const resourceTypesWithoutIndex: string[] = [];
+
+        // eslint-disable-next-line no-restricted-syntax
+        for (const [resourceType, mappings] of Object.entries(this.searchMappings)) {
+            console.log(`sending putMapping request for: ${resourceType}`);
+            try {
+                // eslint-disable-next-line no-await-in-loop
+                await this.searchClient.indices.put_mapping({
+                    index: toIndexName(resourceType),
+                    body: mappings,
+                });
+            } catch (error) {
+                if (error instanceof ResponseError && error.body.error.type === 'index_not_found_exception') {
+                    console.log(`index for ${resourceType} was not found. It will be created`);
+                    resourceTypesWithoutIndex.push(resourceType);
+                } else {
+                    throw error;
+                }
+            }
+        }
+
+        // eslint-disable-next-line no-restricted-syntax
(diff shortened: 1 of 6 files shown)


src/searchMappingsManager/index.ts
```diff
@@ -0,0 +1,77 @@
...
+        for (const [resourceType, mappings] of Object.entries(this.searchMappings)) {
+            console.log(`sending putMapping request for: ${resourceType}`);
+            try {
+                // eslint-disable-next-line no-await-in-loop
+                await this.searchClient.indices.put_mapping({
+                    index: toIndexName(resourceType),
+                    body: mappings,
+                });
+            } catch (error) {
+                if (error instanceof ResponseError && error.body.error.type === 'index_not_found_exception') {
```
Bingjiling: Can we catch the situation where there's conflict in an index, so that we can still create static mapping for everything without a conflict, and log a list of all indexes that have conflict? With the current implementation, the customer will only know the first index that failed mapping update, and once they fix that, they might come back and see another conflict.

src/searchMappingsManager/index.ts
```diff
@@ -0,0 +1,77 @@
...
+
+        // eslint-disable-next-line no-restricted-syntax
+        for (const [resourceType, mappings] of Object.entries(this.searchMappings)) {
+            console.log(`sending putMapping request for: ${resourceType}`);
+            try {
+                // eslint-disable-next-line no-await-in-loop
+                await this.searchClient.indices.put_mapping({
+                    index: toIndexName(resourceType),
+                    body: mappings,
+                });
```
rsmayda: Need to create/worry about Alias' as well?

src/searchMappingsManager/index.ts
```diff
@@ -0,0 +1,77 @@
...
+
+const toIndexName = (resourceType: string) => resourceType.toLowerCase();
+
+// eslint-disable-next-line import/prefer-default-export
+export class SearchMappingsManager {
+    private readonly searchMappings: {
+        [resourceType: string]: any;
+    };
+
+    private readonly numberOfShards: number;
```
rsmayda, line 19: Could you add a comment of where this value should be found/used perhaps add this link to the comment: https://docs.aws.amazon.com/opensearch-service/latest/developerguide/sizing-domains.html#bp-sharding

src/searchMappingsManager/index.ts
carvantes: yeah, that's probably better. I changed it.
As a side effect, the `sls deploy` error message (or CFN deployment error message) now lokks much better:

> An error occurred: UpdateSearchMappingsCustomResource - Received response status [FAILED] from custom resource. Message returned: Failed to update mappings for: ActivityDefinition,Condition,Observation,Procedure,ServiceRequest,CarePlan,Encounter,RiskAssessment, CapabilityStatement,CodeSystem,CompartmentDefinition,ConceptMap,GraphDefinition,ImplementationGuide,MessageDefinition,NamingSystem,OperationDefinition,SearchParameter,StructureDefinition,StructureMap,TerminologyCapabilities,ValueSet,ChargeItem,ChargeItemDefinition,EffectEvidenceSynthesis,EventDefinition,Evidence,EvidenceVariable,ExampleScenario,Group,Invoice,Library,Measure,PlanDefinition,Questionnaire,ResearchDefinition,
ResearchElementDefinition,RiskEvidenceSynthesis,Substance,TestScript (RequestId: 8b74824b-4f0f-4631-9d58-d26d58be2a2f).

src/searchMappingsManager/index.ts
carvantes: I added a link to the docs.

src/searchMappingsManager/index.ts
carvantes: We could create aliases here, but notably there's no straightforward way to create the aliases for multi-tenancy since we don't know the tenantIds ahead of time.

To keep it simple, ddbToES will continue to create the aliases

src/searchMappingsManager/index.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    private readonly searchClient: Client;
+
+    /**
+     * @param options
+     * @param options.searchMappings - search mappings for all FHIR resource types
+     * @param options.numberOfShards - number of shards for each new index created. See the documentation for guidance on how to choose the right number:
+     * https://docs.aws.amazon.com/opensearch-service/latest/developerguide/sizing-domains.html#bp-sharding
```
rsmayda, line 31: I think we want this one line up? Associated with shards parameter

carvantes: afaik this is valid syntax. line breaks are allowed on `@param` descriptions and without a line break it's tedious to read it in a screen. 
IDEs pick up the docs correctly:
Screen Shot 2021-10-08 at 07 18 49

rsmayda: ooooo nice okay thanks for verifying

rsmayda (approved):

PR #129, opened 2021-10-25
feat: enable chained parameters in search

Bingjiling:
Issue #, if available:

Description of changes:

Enable chained parameters in search. 

The chained parameters are process in the following steps: 

1. Verify the chained parameters are valid - chained parameter should point to a valid resource type reference, if there're multiple valid resource types, a single resource type should be specified in search modifier. Chained parameter should return less than 100 hits. 
2. For each chained parameters, search through the chain to get the values used for the main search. e.g.: For ```{{API_URL}}/Patient?general-practitioner:PractitionerRole.location:Location.address-city=NYC``` , we'll search for `Location` whose address-city is NYC(result in `Location/1234`), then search for `PractitionerRole` whose location is Location/1234(result in `PractitionerRole/5678`).
3. Combine results from all chained parameters in a SHOULD(OR) clause
4. Insert it into the "main" search parameters  


carvantes (changes requested):
Please fix first all the lint errors. Otherwise it is hard to review the code.

btw, the eslint version for FWoA packages was recently updated, maybe that's what caused the issue. Make sure that you are using the correct eslint version on your end.

src/QueryBuilder/index.ts
```diff
@@ -165,17 +174,97 @@ function searchRequestQuery(
...
+                        `Chained search parameter '${searchModifier.parameterName}' for resource type ${currentResourceType} points to multiple resource types, please specify.`,
+                    );
+                } else {
+                    organizedChain.push(`${searchModifier.parameterName}:${fhirSearchParam.target[0]}`);
+                    [nextResourceType] = fhirSearchParam.target;
+                }
+
+                currentResourceType = nextResourceType;
+            });
+            return { chain: organizedChain, searchQuery: { [initialSearchParam]: searchValues } };
```
Bingjiling: For `{{API_URL}}/Patient?general-practitioner:PractitionerRole.location.address-city=NYC`, it will return 
```
{
chain: [general-practitioner:PractitionerRole, location:Location]
searchQuery: 
  {
  address-city: 'NYC'
  }
}
```

src/elasticSearchService.ts
```diff
@@ -229,6 +233,58 @@ export class ElasticSearchService implements Search {
...
+                    );
+                }
+                stepQueryParams = {};
+                stepQueryParams[nextStepParam] = hits.map((hit) => `${resourceType}/${hit._source.id}`);
+            }
+            if (chainComplete) {
+                combinedChainedParameters = merge(combinedChainedParameters, stepQueryParams);
+            }
+        }
+        return combinedChainedParameters;
```
Bingjiling, line 293: For `{{API_URL}}/Patient?general-practitioner:PractitionerRole.location:Location.address-city=NYC&organization.name=someName`, it will return
```
{
  'general-practitioner': [ 'PractitionerRole/some-uuid-the-chain-point-to' ],
  organization: [ 'Organization/some-uuid-the-chain-point-to' ]
}
```

src/elasticSearchService.ts
```diff
@@ -229,6 +233,56 @@ export class ElasticSearchService implements Search {
...
+        for (const { chain, searchQuery } of getOrganizedChainedParameters(
+            this.fhirSearchParametersRegistry,
+            request,
+        )) {
+            let stepQueryParams = searchQuery;
+            let chainComplete = true;
+            // eslint-disable-next-line no-restricted-syntax
+            for (const stepParam of chain.slice().reverse()) {
+                const [nextStepParam, resourceType] = stepParam.split(':');
+                const stepReqeust: TypeSearchRequest = { ...request, resourceType, queryParams: stepQueryParams };
```
carvantes: typo `stepReqeust `

src/QueryBuilder/index.ts
carvantes: IMO this structure is not that great as a return value since the `getChainedParametersQuery` method (the only caller of `getOrganizedChainedParameters`) needs to do significant manipulation of this structure before being able to use it for queries.
 
It'd be better to concentrate the parsing logic here and return something that almost maps 1:1 to the query args we need.
Maybe something like: 
```
{
  initialValue: 'NYC',
  chain: [
    {
      resourceType: 'Location',
      searchParam: 'address-city',
    }, 
    {
      resourceType: 'PractitionerRole',
      searchParam: 'location',
    }, 
    {
      resourceType: 'Patient',
      searchParam: 'general-practitioner',
    }
  ]
}
```

src/QueryBuilder/index.ts
```diff
@@ -165,17 +174,97 @@ function searchRequestQuery(
         });
 }
 
+// eslint-disable-next-line import/prefer-default-export
+export const getOrganizedChainedParameters = (
```
carvantes: nit: rename to `parseChainedParameters`. I think that "parsing" is descriptive of what this method is doing.

---

Also, can you move this method to a separate file and add unit tests for it? Similar to `sort.ts`

src/QueryBuilder/index.ts
```diff
@@ -165,17 +174,97 @@ function searchRequestQuery(
...
+    fhirSearchParametersRegistry: FHIRSearchParametersRegistry,
+    request: TypeSearchRequest,
+): any => {
```
carvantes: Do not use `any`, use the concrete type here.
Sometimes you can have implicit types, but in my experience specifying explicitly the return type of functions improves readability by a LOT.

btw, maybe your IDE can auto generate this. In Webstorm it is alt+enter and then "specify type explicitly"

src/QueryBuilder/searchModifiers.ts
```diff
@@ -15,7 +16,9 @@ const parseSearchModifiers = (
     if (modifier.length === 1) {
         return { parameterName: modifier[0], modifier: undefined };
     }
-    modifier[1] = modifier[1].toLowerCase();
+    if (toLowerCase) {
+        modifier[1] = modifier[1].toLowerCase();
+    }
```
carvantes: idk why the `toLowerCase` was there in the first place. Maybe we can just get rid of it instead of adding an extra param

src/elasticSearchService.ts
```diff
@@ -229,6 +233,56 @@ export class ElasticSearchService implements Search {
         }
     }
 
+    // eslint-disable-next-line class-methods-use-this
+    private async getChainedParametersQuery(request: TypeSearchRequest, filters: any[] = []) {
```
carvantes: Specify the return type of this function explicitly

src/elasticSearchService.ts
```diff
@@ -229,6 +233,56 @@ export class ElasticSearchService implements Search {
...
+                const params: Query = {
+                    resourceType,
+                    queryRequest: {
+                        size: MAX_CHAINED_PARAMS_RESULT,
+                        track_total_hits: true,
+                        body: {
+                            query: stepQuery,
+                        },
+                    },
+                };
```
carvantes: use the `fields` parameter so that you only retrieve the `id` field of the matching documents. That should be faster than transmitting the entire documents.

https://www.elastic.co/guide/en/elasticsearch/reference/current/search-fields.html

carvantes (changes requested):

src/QueryBuilder/searchModifiers.ts
Bingjiling: Sure,  I double checked where it's used, it should be fine.

src/elasticSearchService.ts
```diff
@@ -229,6 +234,64 @@ export class ElasticSearchService implements Search {
...
+                    filters,
+                );
+                const params: Query = {
+                    resourceType,
+                    queryRequest: {
+                        size: MAX_CHAINED_PARAMS_RESULT,
+                        track_total_hits: true,
+                        body: {
+                            query: stepQuery,
+                            fields: ['id'],
```
carvantes, line 271: Looks like you need to add `"_source": false` as well. Otherwise you are still getting the entire resource document.

src/QueryBuilder/chain.ts
```diff
@@ -0,0 +1,79 @@
...
+                    searchModifier.parameterName,
+                );
+                if (fhirSearchParam === undefined) {
+                    throw new InvalidSearchParameterError(
+                        `Invalid search parameter '${searchModifier.parameterName}' for resource type ${currentResourceType}`,
+                    );
+                }
+                if (fhirSearchParam.type !== 'reference') {
+                    throw new InvalidSearchParameterError(
+                        `Chained search parameter '${searchModifier.parameterName}' for resource type ${currentResourceType} does not point to another resource.`,
```
rsmayda:
```suggestion
                        `Chained search parameter '${searchModifier.parameterName}' for resource type ${currentResourceType} is not a reference.`,
```

src/QueryBuilder/util.ts
```diff
@@ -0,0 +1,46 @@
...
+        // This may occur if the router has advanced querystring parsing enabled
+        // e.g. {{API_URL}}/Patient?name[key]=Smith may be parsed into {"name":{"key":"Smith"}}
+        throw new InvalidSearchParameterError(`Invalid search parameter: '${searchParameter}'`);
+    });
+
+    return normalizedQueryParams;
+};
+
+export const isChainedParameter = (parameterKey: string) => {
+    return parameterKey.match('[A-Za-z][.][A-Za-z]');
```
rsmayda: `GET Patient?general-practitioner.name=Joe&general-practitioner.address-state=MN`

This regex does work for this chain

src/elasticSearchService.ts
```diff
@@ -229,6 +234,64 @@ export class ElasticSearchService implements Search {
         }
     }
 
+    // Return translated chained parameters that can be used as normal search parameters
```
rsmayda, line 237: Do we follow the below guidance? 

> Note that chained parameters are applied independently to the target resource. For example,
`GET Patient?general-practitioner.name=Joe&general-practitioner.address-state=MN`
may return Patients cared for by Joe from CA and Jane from MN: no one practitioner need satisfy both conditions. E.g. the chains are evaluated separately.

https://www.hl7.org/fhir/search.html#chaining

rsmayda: We should add this as a integ test case

src/constants.ts
```diff
@@ -24,3 +24,5 @@ export const NON_SEARCHABLE_PARAMETERS = [
 ];
 
 export const MAX_ES_WINDOW_SIZE: number = 10000;
+
+export const MAX_CHAINED_PARAMS_RESULT: number = 100;
```
rsmayda, line 28: How would we reach this max, like if some one creates a chain 100 long? NVM I understand more at the bottom

src/elasticSearchService.ts
```diff
@@ -229,6 +234,64 @@ export class ElasticSearchService implements Search {
...
+    // eslint-disable-next-line class-methods-use-this
+    private async getChainedParametersQuery(request: TypeSearchRequest, filters: any[] = []): Promise<{}> {
```
rsmayda, line 239: How does pagination work with chainedParams?

src/QueryBuilder/util.ts
Bingjiling: Emm it can match the `r.n` and `r.a` part for the two parameters right?

src/elasticSearchService.ts
Bingjiling: Yes, we follow the guidance by pushing all chained parameters into a SHOULD(OR) clause when assembling it for the final search query. This scenario is covered by unit test [here](https://github.com/awslabs/fhir-works-on-aws-search-es/pull/129/files#diff-baee28b2589ad3c7198a737b611fd836b6c07600a53b7340e2d32eb7869f9704R994-R997)

src/elasticSearchService.ts
Bingjiling: While searching recursively to 'unchain' the parameters for the main search, we ask for only 100 records and check for the total hits. If total hits is larger than 100, then it fails, if not, then we used the retrieved records to move forward.

carvantes (changes requested):

src/QueryBuilder/util.ts
rsmayda: Ooo gotcha I misunderstood what you were using `match` for it looks like it is just a boolean check -- if so could we use `test` instead:
https://stackoverflow.com/questions/10940137/regex-test-v-s-string-match-to-know-if-a-string-matches-a-regular-expression

src/elasticSearchService.ts
rsmayda: Does pagination work for the entire query itself? i.e. a chained search returns 50 Patients on floor 5 our paging defaults to only return 20 at a time. Can a user use paging token to get the remaining 30?

Bingjiling: Yes, once the entire query is assembled, it works the same way as before with a query that is a bit more complicated. Without chained parameter, the entire query looks like: 
```
bool: {
            filter: additionalFilters,
            must: [mustCondition1, mustCondition2 .....],
        }
``` 

With chained parameter, the query looks like: 
```
bool: {
            filter: additionalFilters,
            must: [
               bool: {
                  should: [chainParamCondition1, chainParamCondition2]
                  }
                  mustCondition1, 
                  mustCondition2 .....],
        }
```

src/elasticSearchService.test.ts
```diff
@@ -1014,6 +1014,7 @@ describe('typeSearch', () => {
                                 _type: '_doc',
                                 _id: 'ab69afd3-39ed-42c3-9f77-8a718a247742_1',
                                 _score: 1,
+                                fields: { id: ['ab69afd3-39ed-42c3-9f77-8a718a247742'] },
                                 _source: {
```
rsmayda, line 1018: Since we are only using `_fields` now do we need this `_source` in our tests? Which return is more accurate from a raw ES search?

carvantes (commented):
Can you also update the README at: 
https://github.com/awslabs/fhir-works-on-aws-search-es/blob/mainline/README.md?plain=1#L102

On the Notes column you can mention any limitations we imposed (e.g. the 100 limit thing)

src/elasticSearchService.test.ts
Bingjiling: It's still needed for the 'main' search to run, which use the `_source` filed. We can use mockOnce many times to have different data returned, but it's a bit of an overkill.

carvantes (approved):

Issue #128, opened 2021-10-21
[Bug] Paged results not returning the same result set


ghost:
**Describe the bug**
When a request results in paging, if you run the same request multiple times the results returned in the set are not always the same.

**To Reproduce**
1. Call request something like {{API_URL}}/tenant/7db7dfb5-2f73-4b96-bcf1-131410518850/Observation?patient=Patient/8743177f-ad8b-4dba-b05d-933419da86df&category=vital-signs&_getpagesoffset=0&_count=20
In our case it was returning 20 of 47.

**Expected behavior**
Results returned should be the same when making the same call.

**Additional context**
If we add a _sort=date or _sort=_lastUpdated to our request does seem to resolve the issue.I have also attached the results of both runs
[Run1.txt](https://github.com/awslabs/fhir-works-on-aws-search-es/files/7392312/Run1.txt)
.
[Run2.txt](https://github.com/awslabs/fhir-works-on-aws-search-es/files/7392314/Run2.txt)

carvantes:
Hi, 

The issue has been fixed by the 3 PRs that are mentioned above. To get the fix you need to update to: 

- fhir-works-on-aws-interface@11.3.0
- fhir-works-on-aws-search-es@3.7.0
- fhir-works-on-aws-routing@6.3.0

At the time of writing this they the latest versions of those packages.

Please try it out. Feel free to reopen this issue if the issue persists.

---

The root cause was something often referred to as the "bouncing results" issue. 
> Imagine that you are sorting your results by a timestamp field, and two documents have the same timestamp. Because search requests are round-robined between all available shard copies, these two documents may be returned in one order when the request is served by the primary, and in another order when served by the replica.
> 
> This is known as the bouncing results problem: every time the user refreshes the page, the results appear in a different order. The problem can be avoided by always using the same shards for the same user, which can be done by setting the preference parameter to an arbitrary string like the user’s session ID.

The fix is to use a `sessionId` as the [Elasticsearch preference parameter](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-search.html#search-preference). The chosen `sessionId` is a hash of the JWT token claims.

- https://www.elastic.co/guide/en/elasticsearch/guide/current/_search_options.html#_preferencehttps://www.elastic.co/guide/en/elasticsearch/reference/current/search-shard-routing.html#shard-and-node-preference
- https://www.elastic.co/guide/en/elasticsearch/guide/current/_search_options.html#_preference

PR #130, opened 2021-10-27
feat: use sessionId as search preference

carvantes:
Issue #, if available: #128

Use the `sessionId` as the [Elasticsearch preference parameter](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-search.html#search-preference) to fix the "bouncing results" issue

- https://www.elastic.co/guide/en/elasticsearch/guide/current/_search_options.html#_preferencehttps://www.elastic.co/guide/en/elasticsearch/reference/current/search-shard-routing.html#shard-and-node-preference
- https://www.elastic.co/guide/en/elasticsearch/guide/current/_search_options.html#_preference

~**NOTE:** build fails until the interface change is released and bumped here.~

Related PR:
- https://github.com/awslabs/fhir-works-on-aws-interface/pull/87
- https://github.com/awslabs/fhir-works-on-aws-routing/pull/140


ssvegaraju (approved):

src/elasticSearchService.ts
```diff
@@ -238,6 +238,7 @@ export class ElasticSearchService implements Search {
             const searchQueryWithAlias = {
                 ...searchQuery.queryRequest,
                 index: getAliasName(searchQuery.resourceType, request.tenantId),
+                ...(request.sessionId && { preference: request.sessionId }),
```
rsmayda, line 241: So `preference` controls which node/shard to choose what is the liklihood of this hash causing a "hot" shard?

carvantes: It shouldn't be an issue.

At scale, with many customers, the load is distributed evenly among shards. Elasticsearch hashes the preference value and uses it to choose an eligible shard (the primary or one of its replicas). The `preference` for a customer changes when they refresh the token (at the very least the JWT `iat` claim changes)

Also, the shard choice from the preference is just a preference and it can be "ignored" if things are failing (although it works the vast majority of the time). e.g. if a node is overwhelmed and it fails or rejects a search request, ES can re-route the query to a different shard.

rsmayda (approved):

Issue #132, opened 2021-11-01
Support SearchParameter with ".extension()" in expression


lquanx:
**Describe the bug**
When I try to make an extension field searchable, I got an error "Unexpected "(". Instead, I was expecting to see one of the following:" Here is the [official example](https://www.hl7.org/fhir/R4/searchparameter-example-extension.json.html) regarding SearchParameter on Extension

**To Reproduce**
1. Create file /implementationGuides/example/SearchParameter.json with [official example](https://www.hl7.org/fhir/R4/searchparameter-example-extension.json.html) 
2. Create /impementationGuides/example/.index.json with content:
```
{
  "index-version": 1,
  "files": [
    {
      "filename": "SearchParameter.json",
      "resourceType": "SearchParameter",
      "id": "CustomSearchParameter",
      "url": "http://fhir.example.com/SearchParameter/CustomSearchParameter",
      "version": "1.0.0",
      "type": "reference"
    }
  ]
}
```
3. run command `yarn run compile-igs`

**Expected behavior**
Compile igs without any errors.

**Versions:**
 -  v2.2.0-smart

carvantes:
Hi @Lin-002,  

The issue is with the FHIRPath expression:
```
DocumentReference.extension('http://example.org/fhir/StructureDefinition/participation-agreement')
```
Specifically, we don't support the [.extension() function](https://www.hl7.org/fhir/fhirpath.html#functions).

The workaround is to use the `.extension.where(url = string)` syntax instead. (`.extension()` is just a syntactical shortcut) i.e.
```
DocumentReference.extension.where(url = 'http://example.org/fhir/StructureDefinition/participation-agreement')
```

We'll evaluate supporting the `.extension()` function, but given that there's a simple workaround it is unlikely to be prioritized soon.
Also the IGs we've seen all use the `.extension.where(url = string)`. Maybe that gives better compatibility since `.extension()` is not truly part of the base [FHIRPath spec](https://hl7.org/fhirpath/), but rather functionality added by FHIR.

Thanks for reporting the issue!

---

Some more context: [FHIRPath](https://hl7.org/fhirpath/) as a whole has some very complex constructs (e.g. projection functions, tree navigation functions, etc.), but search parameters definitions tend to only use a limited set of simple FHIRPath constructs. Our implementation only understands a subset of FHIRPath that is enough to understand most search parameters.

We've tested our implementation with the following IGs

1. US Core (hl7.fhir.us.core)
[...]

PR #134, opened 2021-11-03
feat: support .extension() fn in SearchParameter expressions

carvantes:
Issue #, if available: https://github.com/awslabs/fhir-works-on-aws-search-es/issues/132

All the details are on the above issue. 

This is not blocking the adoption of any IG that we are aware of. However, the official [FHIR example](https://www.hl7.org/fhir/R4/searchparameter-example-extension.json.html) uses the `.extension()` function and not supporting it was causing confusion to customers that wanted to craft their own search parameters.


rsmayda (commented):
Did you have a chance to test this locally, deploy and use the search param?

carvantes:
> Did you have a chance to test this locally, deploy and use the search param?

@rsmayda , initially I did not. Verifying the compiled search param is probably enough. 
However, I just tested it e2e by assembling a small IG with the example search param and deploying. The search param works fine.

rsmayda (approved):

Issue #138, opened 2021-11-09
[Bug] chained parameters search returns all Resources


lquanx:
**Describe the bug**
For API request `GET DetectedIssue?patient:Patient.name=aaa` . I got all DetectedIssue Resource with patient fields

**To Reproduce**
1. Git clone smart-mainline branch ` git clone -b smart-mainline git@github.com:awslabs/fhir-works-on-aws-deployment.git`
2. Deploy AWS FHIR Works
3. Create Patient and DetectedIssue Resources.
```
POST /Patient
{
    "resourceType": "Patient",
    "name": [
        {
            "family": "Chalmers"
        }
    ]
}
```
```
POST /DetectedIssue
{
    "resourceType": "DetectedIssue",
    "patient": {
        "reference": "Patient/{{Patient ID from the above API request}}"
    }
}
```
4. Search DetectedIssue with Patient with any name.
```
GET /DetectedIssue?patient:Patient.name=fdsafds
```
Search DetectedIssue with Patient with empty name
```
GET /DetectedIssue?patient:Patient.name=
```

**Expected behavior**
Empty list of DetectedIssue is returned

**Actual behavior**
A DetectedIssue is returned

lquanx:
Thanks for adding Chained Parameters search. I'm not sure why all resources are returned when I use the chained parameter search though. Thanks for taking a look into the issue.

ssvegaraju:
Hi, thanks for bringing up this issue! I was able to reproduce the behavior described above, and have added this to our backlog. Also, FYI, the behavior with the empty name parameter returning all the resources is expected. However the behavior of a name that doesn't exist returning all of the resources appears to be faulty.

lquanx:
got it. Thanks

ssvegaraju:
Hi @Lin-002, we just published a fix for this in version 3.9.1, thanks again for bringing up this issue. I'm going to close this issue for now, but please feel free to reopen if there's anything else!

PR #139, opened 2021-11-10
fix: chained parameters with zero matches returns empty

ssvegaraju:
Issue #138, if available:

Description of changes:
Update chain param search to no longer return all resources when presented with a nonsense search parameter.

### DIFF
--- a/src/__snapshots__/elasticSearchService.test.ts.snap
+++ b/src/__snapshots__/elasticSearchService.test.ts.snap
@@ -2148,6 +2148,259 @@ Array [
 ]
 `;
 
+exports[`typeSearch query snapshots for chained queryParams with no matches queryParams={"general-practitioner:PractitionerRole.location:Location.address-city":"pwoiejfpow","organization.name":"wefgw"} 1`] = `
+Array [
+  Array [
+    Object {
+      "body": Object {
+        "_source": false,
+        "fields": Array [
+          "id",
+        ],
+        "query": Object {
+          "bool": Object {
+            "filter": Array [
+              Object {
+                "match": Object {
+                  "someFieldThatTellsIfTheResourceIsActive": "AVAILABLE",
+                },
+              },
+            ],
+            "must": Array [
+              Object {
+                "multi_match": Object {
+                  "fields": Array [
+                    "address.city",
+                    "address.city.*",
+                  ],
+                  "lenient": true,
+                  "query": "pwoiejfpow",
+                },
+              },
+            ],
+          },
+        },
+      },
+      "index": "location-alias",
+      "size": 100,
+      "track_total_hits": true,
+    },
+  ],
+  Array [
+    Object {
+      "body": Object {
+        "_source": false,
+        "fields": Array [
+          "id",
+        ],
+        "query": Object {
+          "bool": Object {
+            "filter": Array [
+              Object {
+                "match": Object {
+                  "someFieldThatTellsIfTheResourceIsActive": "AVAILABLE",
+                },
+              },
+            ],
+            "must": Array [
+              Object {
+                "bool": Object {
+                  "should": Array [
+                    Object {
+                      "multi_match": Object {
+                        "fields": Array [
+                          "name",
+                          "name.*",
+                        ],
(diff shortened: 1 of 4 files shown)


src/__snapshots__/elasticSearchService.test.ts.snap
```diff
@@ -2404,6 +2404,418 @@ Array [
...
+                },
+              },
+            ],
+            "must": Array [
+              Object {
+                "bool": Object {
+                  "should": Array [],
+                },
+              },
+            ],
```
carvantes: I think that this query will still match ALL documents.

How was this tested?

src/elasticSearchService.test.ts
```diff
@@ -1137,4 +1137,40 @@ describe('typeSearch', () => {
             ).rejects.toThrowError(InvalidSearchParameterError);
         });
     });
+
+    describe('query snapshots for nonsense chained queryParams', () => {
```
carvantes: Avoid using `nonsense` here. Be specific. Maybe something like "chained parameters with zero matches".

Also, rename the PR.

carvantes (changes requested):

src/elasticSearchService.test.ts
```diff
@@ -1137,4 +1137,40 @@ describe('typeSearch', () => {
...
+                    'organization.name': 'wefgw',
+                },
+            ],
+            [{ 'link:Patient.birthdate': 'ge2020-01-01', 'link:Patient.organization.name': 'opwijeow' }],
+        ]).test('queryParams=%j', async (queryParams: any) => {
+            const fakeSearchResult = {
+                body: {
+                    hits: {
+                        total: {
+                            value: 1,
```
Bingjiling: Should the total hits be 0 if there's no match?

src/elasticSearchService.test.ts
ssvegaraju: Renamed PR and tests.

src/__snapshots__/elasticSearchService.test.ts.snap
ssvegaraju: Looks like there was a bug in the error catching logic that was failing test (and I also had a 1 instead of a 0 in the test).

src/elasticSearchService.test.ts
ssvegaraju: This should be a 0, yes! Also, looks like there was an issue with the index errors not properly being filtered, which should be fixed now.

src/elasticSearchService.ts
```diff
@@ -161,8 +161,25 @@ export class ElasticSearchService implements Search {
...
             const filter = this.getFilters(request);
-
-            const chainedParameterQuery = await this.getChainedParametersQuery(request, filter);
+            const parsedChainParameters = parseChainedParameters(
+                this.fhirSearchParametersRegistry,
+                request.resourceType,
+                request.queryParams,
+            );
+            let chainedParameterQuery;
+            if (parsedChainParameters.length > 0) {
```
carvantes, line 170: Maybe I'm missng something, but I don't see the point of calling `parseChainedParameters` here.

The if `parsedChainParameters.length > 0` check is redundant. When there are no matches the next if `isEmpty(chainedParameterQuery)` will always be true.

It's probably better to revert `getChainedParametersQuery` to accepting `request: TypeSearchRequest` as param. It's simpler.

carvantes:
What happens for `GET /DetectedIssue?patient:Patient.name=`?

ssvegaraju:
> 
> 
> What happens for `GET /DetectedIssue?patient:Patient.name=`?

This returns a bundle of all detected issues with a Patient field. This is similar to the behavior of `GET /Patient?name=` returning a bundle of all Patients.

src/elasticSearchService.ts
ssvegaraju: The reason we had to do it this way is in case a search is made without any chained parameters, in which case the search would just return empty without the check for parsedChainParameters.

carvantes (approved):

PR #142, opened 2021-11-17
chore: release PR, tagging and NPM publish automation

Bingjiling:
Issue #, if available:

Description of changes:

* Run standard-version and create PR on push to mainline with a message that does not start with 'chore'
* Manual trigger action that create a tag and publish to NPM
* Tests done:
  * chore commit does not trigger release PR: https://github.com/Bingjiling/fhir-works-on-aws-search-es/actions/runs/1473749782
  * commit trigger release PR: https://github.com/Bingjiling/fhir-works-on-aws-search-es/runs/4244354038?check_suite_focus=true 
  * Test for NPM publish: https://github.com/Bingjiling/zheyanyu-test-publish-package/runs/4267327228?check_suite_focus=true 

### DIFF
--- a/.github/workflows/release-pr.yml
+++ b/.github/workflows/release-pr.yml
@@ -7,7 +7,11 @@ jobs:
   tag-release:
     name: Create release PR
     runs-on: ubuntu-18.04
-    if: ${{ !startsWith(github.event.head_commit.message, 'chore') }}
+    if: >
+      ${{ startsWith(github.event.head_commit.message, 'feat') ||
+      startsWith(github.event.head_commit.message, 'fix') ||
+      contains(github.event.head_commit.message, 'BREAKING_CHANGE') ||
+      contains(github.event.head_commit.message, '!')}}
     steps:
       - uses: actions/checkout@v2.3.4
         with:
@@ -22,13 +26,13 @@ jobs:
           yarn install
           git config user.name "github-action-automation"
           git config user.email "aws-chamonix-dev@amazon.com"
-          git checkout -b dev-release
+          git checkout -b "dev-release-${{ github.event.head_commit.id }}"
           yarn run standard-version
-          git push --set-upstream origin dev-release
+          git push --set-upstream origin "dev-release-${{ github.event.head_commit.id }}"
       - name: Create Pull Request to merge to mainline
         uses: repo-sync/pull-request@v2
         with:
-          source_branch: dev-release
+          source_branch: "dev-release-${{ github.event.head_commit.id }}"
           destination_branch: mainline
           pr_draft: false
           pr_reviewer: "awslabs/aws-chamonix-fhir"
--- a/.github/workflows/release-pr.yml
+++ b/.github/workflows/release-pr.yml
@@ -1,0 +1,36 @@
+name: create release PR
+on:
+  push:
+    branches:
+      - mainline
+jobs:
+  tag-release:
+    name: Create release PR
+    runs-on: ubuntu-18.04
+    if: ${{ !startsWith(github.event.head_commit.message, 'chore') }}
+    steps:
+      - uses: actions/checkout@v2.3.4
+        with:
+          token: ${{secrets.GITHUB_TOKEN}}
+          fetch-depth: 0
+      - name: Use Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 14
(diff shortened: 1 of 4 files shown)


ssvegaraju (dismissed):
Looks good to me! One question: is npm publish done through a tool? I don't see it being called in tag-and-publish.yml outside of line 23

.github/workflows/tag-and-publish.yml
```diff
@@ -0,0 +1,25 @@
...
+      - name: Tag release
+        run: |
+          yarn --frozen-lockfile
+          yarn run release
+          NEW_VERSION=v$(jq '.version' -r package.json)
+          git ls-remote --exit-code --tags origin $NEW_VERSION || git tag $NEW_VERSION
+          git push origin --tags
+      - uses: JS-DevTools/npm-publish@v1
+        with:
+          token: ${{secrets.GITHUB_TOKEN}}
```
carvantes: Add newline at end of file

.github/workflows/release-pr.yml
```diff
@@ -0,0 +1,36 @@
...
+          git push --set-upstream origin dev-release
+      - name: Create Pull Request to merge to mainline
+        uses: repo-sync/pull-request@v2
+        with:
+          source_branch: dev-release
+          destination_branch: mainline
+          pr_draft: false
+          pr_reviewer: "awslabs/aws-chamonix-fhir"
+          pr_label: "auto-release-pr"
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
carvantes: Add newline at end of file

.github/workflows/tag-and-publish.yml
```diff
@@ -0,0 +1,25 @@
+name: Tag release, publish to NPM
+on: [workflow_dispatch]
+jobs:
+  tag-release:
+    name: Tag release based on CHANGELOG
```
carvantes: This is slightly misleading since the changelog does not determine the release versions, but rather the version in `package.json`

Maybe just rename to "Tag release and publish to npm"

.github/workflows/release-pr.yml
```diff
@@ -0,0 +1,36 @@
+name: create release PR
+on:
+  push:
+    branches:
+      - mainline
+jobs:
+  tag-release:
+    name: Create release PR
+    runs-on: ubuntu-18.04
+    if: ${{ !startsWith(github.event.head_commit.message, 'chore') }}
```
carvantes: There are several commit types that behave like `chore` (they don't increment the version) like `docs`, `refactor`, etc.

https://www.conventionalcommits.org/en/v1.0.0/

.github/workflows/release-pr.yml
```diff
@@ -0,0 +1,36 @@
...
+        with:
+          node-version: 14
+      - name: push change to branch dev-release
+        run: |
+          yarn install
+          git config user.name "github-action-automation"
+          git config user.email "aws-chamonix-dev@amazon.com"
+          git checkout -b dev-release
+          yarn run standard-version
+          git push --set-upstream origin dev-release
```
carvantes: I think that this breaks if there are multiple commits before a release pr is merged.

since a fixed branch name (`dev-release`) is used, the 2nd run of this workflow will fail to `git push` since `dev-release` will be out of sync

Bingjiling:
> Looks good to me! One question: is npm publish done through a tool? I don't see it being called in tag-and-publish.yml outside of line 23

Yes! npm publish is done through `JS-DevTools/npm-publish@v1`, the only input needed is the npm token. I haven't tested it yet, will post the result here after the test.

.github/workflows/release-pr.yml
Bingjiling: Good point. I'll update the branch name to have a random string at the end.

.github/workflows/release-pr.yml
Bingjiling: Good point, I'll update the condition to automatically run when commit message startWith feat or fix or contains BREAKING CHANGE or !, and add a manual trigger for this workflow as well.

.github/workflows/release-pr.yml
Bingjiling: Updated branch name to have the latest commit id that triggered the workflow

carvantes (approved):

PR #145, opened 2021-12-07
fix: boolean token search

Bingjiling:
Issue #, if available:

https://github.com/awslabs/fhir-works-on-aws-deployment/issues/516

Description of changes:

* Update token query to accommodate boolean parameter
* Verified all integ tests passes  


src/QueryBuilder/typeQueries/tokenQuery.test.ts
```diff
@@ -213,6 +217,7 @@ describe('tokenQuery', () => {
                         "identifier.coding.code",
                         "identifier.value",
                         "identifier",
+                        "identifier",
```
rsmayda: Duplicate `identifier` is that right?

src/QueryBuilder/typeQueries/tokenQuery.ts
```diff
@@ -82,7 +82,8 @@ export function tokenQuery(
             `${compiled.path}.code${keywordSuffix}`, // Coding
             `${compiled.path}.coding.code${keywordSuffix}`, // CodeableConcept
             `${compiled.path}.value${keywordSuffix}`, // Identifier, ContactPoint
-            `${compiled.path}${keywordSuffix}`, // code, boolean, uri, string
+            `${compiled.path}${keywordSuffix}`, // code, uri, string
+            `${compiled.path}`, // boolean
```
rsmayda: So we don't want this. Just having `compiled.path` in the array list means it could become a text search, meaning searching ~~'male' could return both 'female' and 'male'.~~ 'fe' could return 'female'.

Based on the `CompiledSearchParam` is there anyway for us to know if a param is a bool or not?

rsmayda (changes requested):

rsmayda (commented):
Also could we add a integ test for this exact use case?

src/QueryBuilder/typeQueries/tokenQuery.ts
Bingjiling: Emmm I'm not getting this one. I tested searching for `gender=female` after this change and it is returning only female records. Compiled search parameter only gives information on the path like `{ resourceType: 'Patient', path: 'gender' }` and `{ resourceType: 'Patient', path: 'active' }`

src/QueryBuilder/typeQueries/tokenQuery.test.ts
Bingjiling: Removed duplicates

src/QueryBuilder/typeQueries/tokenQuery.ts
Bingjiling: Update from earlier discussion: Searching for 'fe' returns no records, which is the expected behavior.

src/QueryBuilder/typeQueries/tokenQuery.ts
```diff
@@ -82,9 +83,15 @@ export function tokenQuery(
             `${compiled.path}.code${keywordSuffix}`, // Coding
             `${compiled.path}.coding.code${keywordSuffix}`, // CodeableConcept
             `${compiled.path}.value${keywordSuffix}`, // Identifier, ContactPoint
-            `${compiled.path}${keywordSuffix}`, // code, boolean, uri, string
+            `${compiled.path}${keywordSuffix}`, // code, uri, string, boolean
         ];
 
+        // accommodate for boolean value when keywordSuffix is used
+        // See https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic-field-mapping.html for more detail
```
carvantes: nit: the comment feels outdated since we now use static mappings instead of dynmaic mappings

carvantes (dismissed):
Looks good. 

Please updated the description to link the GH issue # in the description. It's good for bookkeeping.

src/QueryBuilder/typeQueries/tokenQuery.ts
Bingjiling: Update test cases: Searching for prefix with `-` or `:` separation returns no records 

Tests detail: Create a Patient resource with 

```
"communication": [
          {
            "language": {
              "coding": [
                {
                  "system": "urn:ietf:bcp:47",
                  "code": "es-en", or "es:en"
                  "display": "Spanish"
                }
              ],
              "text": "Spanish"
            }
          }
        ],
```

Search for `language=es` does not return the patients just created. Search for `language=es-en` or `language=es:en` returns the corresponding records.

rsmayda (approved):

PR #147, opened 2021-12-14
chore: add unit test coverage and enforcement

ssvegaraju:
Issue #, if available:

Description of changes:

### DIFF
--- a/.github/workflows/code-coverage.yml
+++ b/.github/workflows/code-coverage.yml
@@ -1,0 +1,17 @@
+#
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#  SPDX-License-Identifier: Apache-2.0
+#
+
+name: Code Coverage Report
+
+on: [push, pull_request]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2 # Check out your repository
+    - run: yarn install # Install dependencies
+    - run: yarn test --coverage # Run test
+    - run: bash <(curl -s https://codecov.io/bash) # Upload to Codecov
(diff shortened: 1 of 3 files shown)


Bingjiling:
The code coverage report was ran twice, one for push and one for pull_request, do we need both triggers?

.github/workflows/code-coverage.yml
```diff
@@ -0,0 +1,17 @@
+#
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#  SPDX-License-Identifier: Apache-2.0
+#
+
+name: Code Coverage Report
+
+on: [push, pull_request]
```
Bingjiling, line 8: The code coverage report was ran twice in this PR, one for push and one for pull_request, do we need both triggers?

.github/workflows/code-coverage.yml
```diff
@@ -0,0 +1,17 @@
...
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2 # Check out your repository
+    - run: yarn install # Install dependencies
+    - run: yarn test --coverage # Run test
```
Bingjiling: Can we use existing script yarn test-coverage here?

.github/workflows/code-coverage.yml
ssvegaraju: Initially it only runs once for the pull request, but it runs again on every push after that I think. Tim mentioned here that it might be good to have both: https://github.com/awslabs/fhir-works-on-aws-deployment/pull/522#discussion_r768154940

Bingjiling (approved):
LGTM

PR #150, opened 2021-12-17
refactor: consolidate parsing logic into FhirQueryParser

carvantes:
The goal of this refactor is to extract the parsing logic so that it can be reused to build something other than Elasticsearch queries(right now we are interested in matching FHIR subscriptions)

The new thing is the `parseSearchQueryParam` that returns a `ParsedFhirQueryParams`

https://github.com/awslabs/fhir-works-on-aws-search-es/blob/7e6d13b04bf2f644d1a4a427c3cf2b8498eed8d5/src/FhirQueryParser/index.ts#L90

https://github.com/awslabs/fhir-works-on-aws-search-es/blob/7e6d13b04bf2f644d1a4a427c3cf2b8498eed8d5/src/FhirQueryParser/index.ts#L83-L86

There are no functional changes. Luckily we have plenty of unit tests in `elasticSearchService.test.ts` that test the entire query building process and facilitate making this refactor. All those tests are unchanged and are passing.

This PR looks big but really there's not that much new code. At a high level the changes are:

- For all the files under `src/QueryBuilder/typeQueries`, the parsing logic was extracted into a new file under `src/FhirQueryParser/typeParsers`. All those files have new unit tests. Tests are mostly the same as (same test cases, both are snapshot tests) except that `QueryBuilder` asserts the shape of the Elasticsearch query and `FhirQueryParser` asserts on the shape of the parsed search values (e.g. `DateSearchValue`)
- Parsing logic from `src/QueryBuilder/index.ts` was moved into `src/FhirQueryParser/index.ts.


ssvegaraju (approved):
Looks good to me

Bingjiling (approved):
LGTM

PR #151, opened 2021-12-17
chore: add logging for unsupported query params

ssvegaraju:
Issue #, if available:
FHIR-223

Description of changes:
Rather than error out calls with these query params, I opted to create routing for these general parameters and simply place a log message that the query param is currently unsupported.


src/constants.ts
```diff
@@ -17,12 +17,13 @@ export const NON_SEARCHABLE_PARAMETERS = [
     SORT_PARAMETER,
     SEARCH_PAGINATION_PARAMS.PAGES_OFFSET,
     SEARCH_PAGINATION_PARAMS.COUNT,
-    '_format',
     '_include',
     '_revinclude',
     ...ITERATIVE_INCLUSION_PARAMETERS,
 ];
 
+export const GENERAL_PARAMETERS = ['_format', '_pretty', '_summary', '_elements'];
```
Bingjiling: Rename it to UNSUPPORTED_PARAMETERS?

ssvegaraju: I named it this based off the FHIR spec: https://www.hl7.org/fhir/http.html#parameters. Should I append UNSUPPORTED to the front instead?

Bingjiling: Yes, please. Appending UNSUPPORTED makes it more readable, or it might create confusion why GENERAL PARAMETERS were disregarded.

src/FhirQueryParser/index.ts
```diff
@@ -182,6 +183,11 @@ export const parseQuery = (
     );
 
     const parsedParams = searchableParams.flatMap(([searchParameter, searchValues]) => {
+        if (UNSUPPORTED_GENERAL_PARAMETERS.includes(searchParameter)) {
+            // since we don't support any of these at the moment, just log a message instead of ignoring and continue.
+            getComponentLogger().info(`Search parameter ${searchParameter} is not currently supported.`);
+            return [];
+        }
```
carvantes: I suggest moving this condition to the `.filter` above. It's a bit confusing to do filtering inside the `flatMap`

it may be better to wait for https://github.com/awslabs/fhir-works-on-aws-search-es/pull/155 to be merged since this PR will have small conflicts with it.

carvantes (changes requested):
The proposed change sounds good to me. It makes sense for search requests like `/Provenance?_count=100&_pretty=true` to succeed instead of failing with `Invalid search parameter '_pretty' for resource type Provenance`. I'm ok with dropping the unsupported params and just logging them.

src/FhirQueryParser/index.ts
ssvegaraju: That makes sense. I also fixed the merge conflicts and updated the PR 👍

src/constants.ts
```diff
@@ -22,6 +22,8 @@ export const NON_SEARCHABLE_PARAMETERS = [
     ...INCLUSION_PARAMETERS,
 ];
 
+export const UNSUPPORTED_GENERAL_PARAMETERS = ['_format', '_pretty', '_summary', '_elements'];
```
carvantes: I think that we need to add `...UNSUPPORTED_GENERAL_PARAMETERS` to the `NON_SEARCHABLE_PARAMETERS` array above(btw ,`_format` is still there).

right now `NON_SEARCHABLE_PARAMETERS` is used in other parts of the code, not just in the parser.

src/FhirQueryParser/index.ts
```diff
@@ -201,6 +202,12 @@ export const parseQuery = (
                 return false;
             }
 
+            if (UNSUPPORTED_GENERAL_PARAMETERS.includes(searchParameter)) {
+                // since we don't support any of these at the moment, just log a message instead of ignoring and continue.
+                getComponentLogger().info(`Search parameter ${searchParameter} is not currently supported.`);
+                return false;
+            }
+
```
carvantes: Let's add those params to `otherParams`

src/constants.ts
ssvegaraju: Added the Unsupported general params array to the non searchable array, and removed _format (probably missed from resolving the merge conflicts), thanks for catching that!

src/FhirQueryParser/index.ts
```diff
@@ -201,6 +202,13 @@ export const parseQuery = (
                 return false;
             }
 
+            if (UNSUPPORTED_GENERAL_PARAMETERS.includes(searchParameter)) {
```
carvantes: with the latest changes this condition needs to be moved before the `if(NON_SEARCHABLE_PARAMETERS.includes(searchParameter)) {`at L200. Otherwise this code is unreachable.

carvantes (changes requested):

src/FhirQueryParser/index.ts
ssvegaraju: Thank you for catching that. Just updated it.

carvantes (approved):

PR #153, opened 2022-01-10
feat: add method validateSubscriptionSearchCriteria 

Bingjiling:
Issue #, if available:

Description of changes:

* Add method validateSubscriptionSearchCriteria 
* Update/Add unit tests 

### DIFF
--- a/src/FhirQueryParser/index.test.ts
+++ b/src/FhirQueryParser/index.test.ts
@@ -76,17 +76,9 @@ describe('queryParser', () => {
     });
 
     test('string AND', () => {
-        const q = parseQuery(fhirSearchParametersRegistry, 'Patient', {
-            'name:exact': ['John', 'Anna'],
-            'organization.name': 'HL7',
-        });
+        const q = parseQuery(fhirSearchParametersRegistry, 'Patient', { 'name:exact': ['John', 'Anna'] });
         expect(q).toMatchInlineSnapshot(`
             Object {
-              "chainedSearchParams": Object {
-                "organization.name": Array [
-                  "HL7",
-                ],
-              },
               "resourceType": "Patient",
               "searchParams": Array [
                 Object {
@@ -176,17 +168,9 @@ describe('queryParser', () => {
     });
 
     test('date', () => {
-        const q = parseQuery(fhirSearchParametersRegistry, 'Patient', {
-            birthdate: '1999-09-09',
-            _include: 'Patient:someField',
-        });
+        const q = parseQuery(fhirSearchParametersRegistry, 'Patient', { birthdate: '1999-09-09' });
         expect(q).toMatchInlineSnapshot(`
             Object {
-              "inclusionSearchParams": Object {
-                "_include": Array [
-                  "Patient:someField",
-                ],
-              },
               "resourceType": "Patient",
               "searchParams": Array [
                 Object {
@@ -310,17 +294,9 @@ describe('queryParser', () => {
     });
 
     test('token', () => {
-        const q = parseQuery(fhirSearchParametersRegistry, 'Patient', {
-            identifier: 'http://acme.org/patient|2345',
-            _count: '20',
-        });
+        const q = parseQuery(fhirSearchParametersRegistry, 'Patient', { identifier: 'http://acme.org/patient|2345' });
         expect(q).toMatchInlineSnapshot(`
             Object {
-              "otherParams": Object {
-                "_count": Array [
-                  "20",
-                ],
-              },
               "resourceType": "Patient",
               "searchParams": Array [
                 Object {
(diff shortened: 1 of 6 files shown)


src/FhirQueryParser/index.ts
```diff
@@ -202,5 +203,6 @@ export const parseQuery = (
     return {
         resourceType,
         searchParams: parsedParams,
+        hasUnsearchableParams: size(normalizedQueryParams) > searchableParams.length,
```
carvantes: I think that this does not belong here. I think it's best to avoid having subscription-specific logic on the parser.

It's not necessarily true that this params are "unsearchable". They may searchable, but just not usable in subscriptions due to their complexity. I'd suggest moving this check outside of this method.

src/elasticSearchService.test.ts
```diff
@@ -1174,3 +1174,51 @@ describe('typeSearch', () => {
...
+        );
+    });
+
+    describe('Valid search string', () => {
+        each(['Patient?name=Harry', 'Patient?name=Harry&family=Potter']).test('queryString=%j', (queryString: any) => {
+            const es = new ElasticSearchService(FILTER_RULES_FOR_ACTIVE_RESOURCES);
+
+            expect(() => {
+                es.validateSubscriptionSearchCriteria(queryString);
+            }).toBeDefined();
```
carvantes: This assertion is not really doing anything. `validateSubscriptionSearchCriteria` is not being called. `expect` is just asserting that the arrow function is defined.

e.g. this passes:
```
expect(() => { throw new Error(); }).toBeDefined();
```
 use `expect().not.toThrow()` instead

src/elasticSearchService.ts
```diff
@@ -519,6 +521,27 @@ export class ElasticSearchService implements Search {
...
     }
 
+    validateSubscriptionSearchCriteria(searchCriteria: string): void {
+        if (!searchCriteria.includes('?')) {
+            throw new InvalidSearchParameterError(
+                'Search string used for field criteria does not contain ?, please use valid search string',
+            );
+        }
+        const [resourceType, ...queryStrings] = searchCriteria.split('?');
+        const queryString = queryStrings.join('?');
```
carvantes: I think we can clean up this a little bit. `queryStrings` is not truly query strings. It may be just a bunch of text fragments since `?` beyond the first are taken literally as strings . (e.g. `Patient?name=Anne?as?&phone=333?222`)

code would be easier to follow if you find the index of the first `?` and then slice on it.

Also, can you add a unit test with multiple `?` on the query params?

carvantes (changes requested):

src/FhirQueryParser/index.ts
Bingjiling: Yes, I agree subscription logic should not interfere with existing search logic as much as possible. I tried doing it outside of this method first, but it turns out to be messy as we'll be picking apart the `parseQuery` function to give the "interim result" needed by the validation. I think this way is actually less destruction with the parser function instead of breaking it into multiple pieces. 

I can update the parameter name to be `hasUnsearchableOrChainedParams`?

src/elasticSearchService.test.ts
Bingjiling: Good catch!

src/FhirQueryParser/index.ts
carvantes: I think that we should strive to keep the parser clean. We can improve the parser to make this easier though. Take a look at https://github.com/awslabs/fhir-works-on-aws-search-es/pull/155
Once we merge that PR you can validate the subscription outside of the parser by checking that 
`parsed.inclusionSearchParams`, `parsed.chainedSearchParams` and `parsed.otherParams` are all empty.

src/FhirQueryParser/index.test.ts
```diff
@@ -168,9 +176,17 @@ describe('queryParser', () => {
     });
 
     test('date', () => {
-        const q = parseQuery(fhirSearchParametersRegistry, 'Patient', { birthdate: '1999-09-09' });
+        const q = parseQuery(fhirSearchParametersRegistry, 'Patient', {
+            birthdate: '1999-09-09',
+            _include: 'Patient:someField',
```
carvantes: are the changes in this test file necessary?

We already have dedicated tests for chained, inclusion and other params near the end of this file. They are not wrong, but I think it's confusing to add something like an `_include` param here when this test is meant to be focused on the `date` params.

carvantes (commented):
Looking good. 

I think that we should support the case of subscribing to changes on ANY resource of a given resourceType. e.g. the criteria for ALL patients would be `Patient` or `Patient?`. 

It looks like you only need some minor changes in parsing (`?` may not be mandatory) to support this case.

carvantes (approved):

Issue #154, opened 2022-01-10
[Bug] revInclude on a resource only returns a max of 10 records


ghost:
**Describe the bug**
When running a revInclude for Provenance on a MedicationRequest Resource it is only returning 10 provenance records when there are more than that tied to it.

**To Reproduce**
Example:
https://{{API_URL}}/tenant/7db7dfb5-2f73-4b96-bcf1-131410518850/MedicationRequest?patient=Patient/cc8d11df-2df0-4c0a-b478-8169a9228ae3&intent=order&_revinclude=Provenance:target&_count=20

**Expected behavior**
The expectation is that it would return all provenance records for the MedicationRequest resource.

ssvegaraju:
Hi @medhost-gkersting,
I was working on finding where this bug was happening, but noticed this morning that the query above (which had correctly reproduced the bug before) was working correctly. I just wanted to check in with you to see if this was still an issue you were experiencing?

Thanks for your time,
Sukeerth

ghost:
@ssvegaraju We will need to pull in some of the latest changes and see if we can still reproduce the issue.

ssvegaraju:
Hi @medhost-gkersting, 
Just wanted to update you on this and let you know that we have merged a fix into mainline for this bug, and it should be available with search-es package v3.9.3 shortly. Thanks again for bringing it to our attention! I'll go ahead and close the issue for now, but feel free to reopen if needed.

Best,
Sukeerth

PR #155, opened 2022-01-12
refactor: add chained and inclusion params to parser

carvantes:
It's better for the parser to report include/revinclude, chained params, etc. instead of silently dropping them.

Eventually we could consolidate all the parsing here (e.g. move [parseChainedParameters](https://github.com/awslabs/fhir-works-on-aws-search-es/blob/3a134092705ef9a44db8feaea58dd2d00d854f20/src/QueryBuilder/chain.ts#L12) here), but right now the parser just reports a list of the include/revinclude, chained params, and other non searchable params.

### DIFF
--- a/src/FhirQueryParser/index.test.ts
+++ b/src/FhirQueryParser/index.test.ts
@@ -328,4 +328,63 @@ describe('queryParser', () => {
             }
         `);
     });
+
+    test('chained params', () => {
+        const q = parseQuery(fhirSearchParametersRegistry, 'DiagnosticReport', {
+            'subject.name': 'DiagnosticReport?subject.name=peter',
+        });
+        expect(q).toMatchInlineSnapshot(`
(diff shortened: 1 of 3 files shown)


ssvegaraju (approved):
Looks good to me. Does this change have any effect on #154 (the bug with _revinclude only returning a max of 10 items)? I'm currently working on a fix for that so I just wanted to double check.

carvantes:
> Looks good to me. Does this change have any effect on #154 (the bug with _revinclude only returning a max of 10 items)? I'm currently working on a fix for that so I just wanted to double check.

afaik it deoesn't affect that bug. In general this shouldn't change anything since the new fields(e.g. `inclusionSearchParams`) are not being used at all yet (they'll make subscriptions easier later). But do make sure that your fix works well on top of this PR

PR #160, opened 2022-01-21
refactor: add referenceParser

carvantes:
Extracted the reference parsing logic into `referenceParser.ts`. There are no functional changes.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #161, opened 2022-01-24
feat: add referenceMatcher

carvantes:
Adds the referebce matcher to the InMemoryMatcher

This PR builds on top on recent changes to the referenceParser:https://github.com/awslabs/fhir-works-on-aws-search-es/pull/160

### DIFF
--- a/src/InMemoryMatcher/index.ts
+++ b/src/InMemoryMatcher/index.ts
@@ -18,12 +18,15 @@ import { dateMatch } from './matchers/dateMatch';
 import { getAllValuesForFHIRPath } from '../getAllValuesForFHIRPath';
 import { stringMatch } from './matchers/stringMatch';
 import { quantityMatch } from './matchers/quantityMatch';
+import { referenceMatch } from './matchers/referenceMatcher';
+import { ReferenceSearchValue } from '../FhirQueryParser/typeParsers/referenceParser';
 
 const typeMatcher = (
     searchParam: SearchParam,
     compiledSearchParam: CompiledSearchParam,
     searchValue: unknown,
     resourceValue: any,
+    { baseUrl }: { baseUrl?: string } = {},
 ): boolean => {
     switch (searchParam.type) {
         case 'string':
@@ -35,7 +38,10 @@ const typeMatcher = (
         case 'quantity':
             return quantityMatch(searchValue as QuantitySearchValue, resourceValue);
         case 'reference':
-            break;
+            return referenceMatch(searchValue as ReferenceSearchValue, resourceValue, {
+                target: searchParam.target,
+                baseUrl,
+            });
         case 'token':
             break;
         case 'composite':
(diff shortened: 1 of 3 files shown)


src/InMemoryMatcher/index.ts
```diff
@@ -18,12 +18,15 @@ import { dateMatch } from './matchers/dateMatch';
...
 import { quantityMatch } from './matchers/quantityMatch';
+import { referenceMatch } from './matchers/referenceMatcher';
+import { ReferenceSearchValue } from '../FhirQueryParser/typeParsers/referenceParser';
 
 const typeMatcher = (
     searchParam: SearchParam,
     compiledSearchParam: CompiledSearchParam,
     searchValue: unknown,
     resourceValue: any,
+    { baseUrl }: { baseUrl?: string } = {},
```
Bingjiling: Why is baseURL optional here?

src/InMemoryMatcher/matchers/referenceMatcher.test.ts
```diff
@@ -0,0 +1,110 @@
...
+                    target: [],
+                }),
+            ).toBe(false);
+        });
+    });
+
+    describe('url', () => {
+        test('relative reference in resource with matching baseUrl', () => {
+            const searchValue: ReferenceSearchValue = {
+                fhirServiceBaseUrl: 'https://fwoa.com',
```
Bingjiling, line 43: The naming here `fhirServiceBaseUrl` v.s. `baseUrl` is a bit confusing.

src/InMemoryMatcher/matchers/referenceMatcher.test.ts
```diff
@@ -0,0 +1,110 @@
...
+            ).toBe(true);
+        });
+
+        test('matching full url in resource', () => {
+            const searchValue: ReferenceSearchValue = { id: '111', referenceType: 'idOnly' };
+            const resourceValue = { reference: 'https://fwoa.com/Patient/111' };
+            expect(
+                referenceMatch(searchValue, resourceValue, {
+                    baseUrl: 'https://fwoa.com',
+                    target: ['Patient'],
```
Bingjiling: Add a test case where there're multiple items in target?

carvantes: I added more items to the test array.

src/InMemoryMatcher/matchers/referenceMatcher.test.ts
carvantes: I renamed so that both are `fhirServiceBaseUrl` since they have the same meaning and added docs to describe the fn parameters.

src/InMemoryMatcher/index.ts
carvantes: This gives more flexibility. `baseUrl` is only used by the `referenceMatch` and is also an optional param there. The idea is that if baseUrl is missing the matcher works but it won't translate relative references into full urls before matching.

For the subscription use case we'll always pass a `baseUrl`, but the matcher itself can work without it.

Bingjiling (approved):

PR #164, opened 2022-01-31
fix: allow revinclude to return more than 10 resources

ssvegaraju:
Issue #154 , if available:

Description of changes:
Added a size parameter to the multisearch used by the inclusion parameters (include and revinclude). Updated test snapshots to include the size parameter.


src/elasticSearchService.ts
```diff
@@ -367,7 +367,7 @@ export class ElasticSearchService implements Search {
         const apiResponse = await this.esClient.msearch({
             body: searchQueriesWithAlias.flatMap((query) => [
                 { index: query.index, ...(request.sessionId && { preference: request.sessionId }) },
-                { query: query.body!.query },
+                { query: query.body!.query, ...{ size: 1000 } },
```
carvantes: Create a constant for the `1000`

Also, I think that we should make this change somewhere else. This is the `executeQueries` method and it is misleading that it is modifying queries before executing them. I think that we should make the change where the queries are being built. e.g. https://github.com/awslabs/fhir-works-on-aws-search-es/blob/9847e1415d5067496b391e733fdd365185a062f0/src/searchInclusions.ts#L175

carvantes (changes requested):
Solution is correct. I left a couple of comments to tidy up the implementation.

Also, consider renaming the PR. Something like "fix: allow revinclude to return more than 10 resources". 
It's often better to think about the end result instead of the specific internals being changed when thinking of a PR name.

src/elasticSearchService.ts
ssvegaraju: I implemented `size` into the inclusion queries, and double checked that any queries made without a `size` parameter would still work as before with the changes made to `executeQueries()`. Not sure if MAX_NUM_ES_HITS is an appropriate name for the constant, however.

src/constants.ts
```diff
@@ -26,3 +26,5 @@ export const NON_SEARCHABLE_PARAMETERS = [
 export const MAX_ES_WINDOW_SIZE: number = 10000;
 
 export const MAX_CHAINED_PARAMS_RESULT: number = 100;
+
+export const MAX_NUM_ES_HITS: number = 1000;
```
carvantes: this const is specific to inclusion params. Maybe rename to `MAX_INCLUSION_PARAM_RESULTS`

src/elasticSearchService.ts
carvantes: Looks good. I just left a minor comment with a name suggestion for the constant

src/constants.ts
ssvegaraju: Thank you for the name suggestion, I've updated it as such!

carvantes (approved):

Issue #166, opened 2022-02-02
Chained search parameters should resolve to exactly one resource type


ghost:
**Describe the bug**
Some chained search parameter searches are erroring with '...points to multiple resource types, please specify'.  FHIR specifies that chained params must resolve to exactly one resourceType but the currently the code only looks at the target array when deciding if chained parameters are ambiguous

**To Reproduce**
Perform a DocumentReference search request: /DocumentReference?patient.identifier=2.16.840.1.113883.3.1579|8889154591540

**Expected behavior**
a search against the DocumentReference patient identifier

**Additional context**
The search param definition has "target" : ["Patient", "Group"] but its expression is “DocumentReference.subject.where(resolve() is Patient)”. Logically this means that the search parameter can only match Patient

nguyen102:
Thanks for the bug report. We're currently working on a fix for this bug and will keep this issue updated with our progress.

ssvegaraju:
Hi, we have just merged a fix for this issue into mainline, and it should be available with search-es v3.9.4!

PR #168, opened 2022-02-03
fix: chain parameters should inspect conditions to narrow down possible target types

ssvegaraju:
Issue #166, if available:

Description of changes:
Added logic for parsing compiledSearchParamaters' conditions to narrow down target Resource types when there are multiple, if possible.

### DIFF
--- a/src/QueryBuilder/chain.test.ts
+++ b/src/QueryBuilder/chain.test.ts
@@ -161,6 +161,26 @@ describe('parseChainedParameters', () => {
                     },
                 ],
             },
+            {
+                name: 'patient',
+                url: 'http://hl7.org/fhir/SearchParameter/Person-patient',
+                type: 'reference',
+                description: 'The Person links to this Patient',
+                base: 'Person',
+                target: ['Patient', 'Practitioner'],
+                compiled: [
+                    {
+                        resourceType: 'Person',
+                        path: 'link.target',
+                        condition: ['link.target', 'resolve', 'Observation'],
+                    },
+                    {
+                        resourceType: 'Person',
+                        path: 'link.target',
+                        condition: ['link.something', 'resolve', 'Observation'],
+                    },
+                ],
+            },
         ];
 
         const successCase: SearchParam = {
@@ -185,6 +205,6 @@ describe('parseChainedParameters', () => {
         };
 
         expect(getUniqueTarget(successCase)).toMatchInlineSnapshot(`"Patient"`);
-        errorCases.forEach((errorCase) => expect(getUniqueTarget(errorCase)).toMatchInlineSnapshot(`undefined`));
+        errorCases.forEach((errorCase) => expect(getUniqueTarget(errorCase)).toBeUndefined());
     });
 });
--- a/src/QueryBuilder/chain.test.ts
+++ b/src/QueryBuilder/chain.test.ts
@@ -62,6 +62,30 @@ describe('parseChainedParameters', () => {
               },
             ]
         `);
+
+        expect(
+            parseChainedParameters(fhirSearchParametersRegistry, 'DocumentReference', {
+                'patient.identifier': '2.16.840.1.113883.3.1579|8889154591540',
+            }),
+        ).toMatchInlineSnapshot(`
+            Array [
+              Object {
+                "chain": Array [
+                  Object {
+                    "resourceType": "Patient",
+                    "searchParam": "identifier",
+                  },
+                  Object {
+                    "resourceType": "DocumentReference",
+                    "searchParam": "patient",
+                  },
+                ],
+                "initialValue": Array [
+                  "2.16.840.1.113883.3.1579|8889154591540",
+                ],
+              },
+            ]
+        `);
     });
 
     test('invalid params', () => {
--- a/src/QueryBuilder/chain.test.ts
+++ b/src/QueryBuilder/chain.test.ts
@@ -4,8 +4,8 @@
  */
 
 import { InvalidSearchParameterError } from 'fhir-works-on-aws-interface';
-import parseChainedParameters from './chain';
-import { FHIRSearchParametersRegistry } from '../FHIRSearchParametersRegistry';
+import parseChainedParameters, { getUniqueTarget } from './chain';
+import { FHIRSearchParametersRegistry, SearchParam } from '../FHIRSearchParametersRegistry';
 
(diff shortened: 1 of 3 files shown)


src/QueryBuilder/chain.ts
```diff
@@ -55,9 +55,28 @@ const parseChainedParameters = (
                         );
                     }
                 } else if (fhirSearchParam.target?.length !== 1) {
-                    throw new InvalidSearchParameterError(
-                        `Chained search parameter '${searchModifier.parameterName}' for resource type ${currentResourceType} points to multiple resource types, please specify.`,
-                    );
+                    // check compiled[].condition for resolution
+                    if (fhirSearchParam.compiled.length > 0) {
+                        const compiled = fhirSearchParam.compiled.pop()!; // we can use ! since we checked length before
```
carvantes: I don't think it's safe to use `pop` here since it mutates the array. `fhirSearchParam` comes from `fhirSearchParametersRegistry` and calling `pop` will permanently remove `compiled` items from the array and cause trouble in future requests.

src/QueryBuilder/chain.test.ts
```diff
@@ -62,6 +62,30 @@ describe('parseChainedParameters', () => {
               },
             ]
         `);
+
+        expect(
+            parseChainedParameters(fhirSearchParametersRegistry, 'DocumentReference', {
```
carvantes: move this to a separate test. Name it something like "search param with conditions that narrow down target type"

src/QueryBuilder/chain.ts
```diff
@@ -55,9 +55,28 @@ const parseChainedParameters = (
...
+                        // if there is no resolve condition, we have multiple resources pointed to.
+                        // condition's format is defined in `../FHIRSearchParamtersRegistry/index.ts`
+                        if (!compiled.condition || compiled.condition[1] !== COMPILED_CONDITION_OPERATOR_RESOLVE) {
+                            throw new InvalidSearchParameterError(
+                                `Chained search parameter '${searchModifier.parameterName}' for resource type ${currentResourceType} points to multiple resource types, please specify.`,
```
carvantes: Refactor so that this error is thrown only once. It makes updating the error message later easier.

In general it'd be better to move most of the new code in this PR to a separate function. Maybe something like `getUniqueTarget(fhirSearchParam): string`

src/QueryBuilder/chain.ts
carvantes: What happens if there are multiple `compiled` ? 
I think that we should only narrow down the type after verifying that ALL `compiled` have a condition that resolves to a type.
Sorry if my earlier slack message wasn't clear.

carvantes (changes requested):

src/QueryBuilder/chain.test.ts
ssvegaraju: I also added tests for the edge cases we went over!

src/QueryBuilder/chain.ts
ssvegaraju: Thanks for helping clear up the confusion on the edge cases, I refactored to a method and removed the mutation of `fhirSearchParametersRegistry`

src/QueryBuilder/chain.ts
```diff
@@ -1,14 +1,36 @@
...
+import { FHIRSearchParametersRegistry, SearchParam } from '../FHIRSearchParametersRegistry';
+import { COMPILED_CONDITION_OPERATOR_RESOLVE, NON_SEARCHABLE_PARAMETERS } from '../constants';
 import { parseSearchModifiers, normalizeQueryParams, isChainedParameter } from '../FhirQueryParser/util';
 
 export interface ChainParameter {
     chain: { resourceType: string; searchParam: string }[];
     initialValue: string[];
 }
 
+export function getUniqueTarget(fhirSearchParam: SearchParam): string | undefined {
```
carvantes, line 12: Can you move the rest of the existing logic regarding finding target into this function?
I think that this function should stand on its own and give us the unique target no matter what. Right now the function assumes that the `fhirSearchParam.target?.length !== 1` check was done outside.

src/QueryBuilder/chain.test.ts
```diff
@@ -93,4 +93,98 @@ describe('parseChainedParameters', () => {
...
+                {
+                    resourceType: 'Person',
+                    path: 'link.target',
+                    condition: ['link.something', 'resolve', 'Patient'],
+                },
+            ],
+        };
+
+        expect(getUniqueTarget(successCase)).toMatchInlineSnapshot(`"Patient"`);
+        errorCases.forEach((errorCase) => expect(getUniqueTarget(errorCase)).toMatchInlineSnapshot(`undefined`));
```
carvantes: Use `.toBeUndefined()` instead. 

`toMatchInlineSnapshot` should not be used inside loops(since it needs to write potentially different snapshots for each tests). I think that In this particular case it works since they all have the same snapshot and you wrote the snapshot manually. AFAIK Jest crashes if you attempt to update that snapshot in a loop

src/QueryBuilder/chain.ts
carvantes: consider the case where the conditions resolve to a type that is NOT in the target array(It's unlikely since that means the search parameter definition is messed up). It should return undefined in that case. e.g:
```json
{
    "name": "patient",
    "url": "http://hl7.org/fhir/SearchParameter/aaa",
    "type": "reference",
    "description": "...",
    "base": "Person",
    "target": [
      "Patient",
      "Practitioner"
    ],
    "compiled": [
      {
        "resourceType": "Person",
        "path": "link.target",
        "condition": [
          "link.target",
          "resolve",
          "Observation"
        ]
      },
      {
        "resourceType": "Person",
        "path": "link.target",
        "condition": [
          "link.something",
          "resolve",
          "Observation"
        ]
      }
    ]
  },
```

carvantes (changes requested):

src/QueryBuilder/chain.ts
ssvegaraju: Added this new edge case, and pulled out the if statement in parseChainedParameters.

src/QueryBuilder/chain.test.ts
ssvegaraju: I did not know that about jest, thanks for letting me know! Amended the test cases to use toBeUndefined.

carvantes (approved):

PR #182, opened 2022-03-14
feat: Validate inclusion params before searching

Bingjiling:
Issue #, if available:

Description of changes:

* Refactor inclusion logic so inclusion search param is parsed under FhirQueryParser
* Wildcard inclusion search param is parsed after main search is run as main search outcome is needed for parsing wildcard param
* Strictly validate inclusion search params 
* Valid inclusion search params are treated exactly the same as before 
* Deployed and tested in AWS Account 

### DIFF
--- a/src/elasticSearchService.test.ts
+++ b/src/elasticSearchService.test.ts
@@ -506,7 +506,6 @@ describe('typeSearch', () => {
             [{ _revinclude: '*' }],
             [{ _revinclude: 'MedicationAdministration:request' }],
             [{ _revinclude: 'MedicationAdministration:request:MedicationRequest' }],
-            // [{ _revinclude: 'MedicationAdministration:request:Device' }],
             [{ _revinclude: ['MedicationAdministration:request', 'Provenance:target'] }],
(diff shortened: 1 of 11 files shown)


src/elasticSearchService.test.ts
```diff
@@ -506,7 +506,6 @@ describe('typeSearch', () => {
             [{ _revinclude: '*' }],
             [{ _revinclude: 'MedicationAdministration:request' }],
             [{ _revinclude: 'MedicationAdministration:request:MedicationRequest' }],
-            [{ _revinclude: 'MedicationAdministration:request:Device' }],
```
ssvegaraju, line 509: Is there a reason we are removing this test?

Bingjiling: Yes, with the updated logic this test will fail since the MedicationAdministration:request:Device is not a valid. Before the change, invalid parameter was simply ignored. 

For invalid inclusion parameters, tests were added in the new test file, so this one was removed.

ssvegaraju (approved):

